PR #1, opened 2021-02-08
[EDNA-4] Init repo

gromakovsky:
## Description

This PR initiates this repo based on the template it was created from.

**TODO**: pick a license and update everywhere, ensure correct licensing information. We want AGPL, need a confirmation. It will be done separately.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-4

- [x] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).


pva701 (approved):
nothing special

pva701:
@gromakovsky shouldn't we prepend commit messages with issue tag?

gromakovsky:
> @gromakovsky shouldn't we prepend commit messages with issue tag?

I was thinking about that. Since we want this tool to be used by a broad audience and be a useful open-source tool, I think including IDs of private issues (that only we can see) doesn't make much sense. As a middle ground, I propose to mention issues in PRs and not mention them in commit messages. Since ≈all changes come from PRs, it should be easy to figure out to which issue a commit refers.

sashasashasasha151 (approved):

PR #3, opened 2021-02-11
Initial SQL schema

pva701:
[//]: # (This is a template of a pull request template.)
[//]: # (You should modify it considering specifics of a particular repository and put it there.)
[//]: # (Comments like this are meta-comments, they shouldn't be present in the final template.)
[//]: # (Comments starting with '

[//]: # (Update link to style guide if necesary or remove if it's not present)

- [x] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).

### DIFF
--- a/backend/sql/initial_schema.sql
+++ b/backend/sql/initial_schema.sql
@@ -1,0 +1,104 @@
+create table project
+(
+    project_id  serial not null primary key,
+    name        text   not null unique,
+    description text   null
+);
+
+create table test_methodology
+(
+    test_methodology_id serial        not null primary key,
+    name                text          not null unique,
+    description         text          null,
+    confluence_link     varchar(1000) null
+);
+
+-- Corresponds to an EXCEL file
+create table experiment
+(
+    experiment_id  serial                   not null primary key,
+    project_id     int                      not null,
+    methodology_id int                      not null,
+    -- This is input from the user
+    name           text                     null,
+    meta           json                     null,
+    -- This is fallback name to show if name is not specified
+    raw_file_name  text                     not null,
+    upload_date    timestamp with time zone not null,
+
+    constraint belongs_to_project
+        foreign key (project_id) references project (project_id)
+            on update no action
+            on delete cascade,
+
+    constraint has_methodology
+        foreign key (methodology_id) references test_methodology (test_methodology_id)
+            -- Methodology_id is immutable
+            on update no action
+            -- Prevent removing methodology if there is an experiment having such methodology
+            on delete no action
+);
+
+create table measurement
+(
+    measurement_id serial           not null primary key,
+    experiment_id  int              not null,
+    -- Either compound id or name
+    compound_id    text             not null,
+    concentration  double precision not null,
+    signal         double precision not null,
+
+    constraint belongs_to_experiment
+        foreign key (experiment_id) references experiment (experiment_id)
+            on delete cascade
+            on update cascade
+);
+
+create table analysis_method
+(
+    analysis_method_id serial not null primary key,
+    description        text   null,
+    -- Analysis-specific parameters
+    parameters         json   null
+);
+
+create table analysis
+(
+    analysis_id        serial not null primary key,
+    analysis_method_id int    not null,
+    experiment_id      int    not null,
+    compound_id        int    not null,
+    is_suspicious      bool   not null,
+    -- Maybe JSON better?
+    result             json   not null,
+
+    constraint determined_by_analysis_method
+        foreign key (analysis_method_id) references analysis_method (analysis_method_id)
+            on delete cascade
+            on update cascade,
+
+    constraint such_experiment_exists
+        foreign key (experiment_id) references experiment (experiment_id)
+            on delete cascade
+            on update cascade
+
+    -- TODO do we want to make sure that there is at least one measurement for (experiment_id, compound_id) in
+    -- measurement table?
+);
+
+create table analysis_removed_measurements
+(
+    analysis_id    int not null,
+    measurement_id int not null,
+    primary key (analysis_id, measurement_id),
+
+    constraint belongs_to_analysis
+        foreign key (analysis_id) references analysis (analysis_id)
+            on delete cascade
+            on update cascade,
+
+    constraint such_measurement_exists
+        foreign key (measurement_id) references measurement (measurement_id)
+            on delete cascade
+            on update cascade
+);
--- a/README.md
+++ b/README.md
@@ -20,7 +20,8 @@ There is nothing to build yet.
 
 ## Usage
 
-There is nothing to use yet.
+To run a postgres container run `docker-compose up postgres` in `deployment` directory,
+then you can connect to it via `psql` or using any convenient database viewer.
 
 ## Issue Tracker
 
--- a/backend/sql/drop.sql
+++ b/backend/sql/drop.sql
@@ -1,0 +1,7 @@
+drop table if exists project cascade;
+drop table if exists test_methodology cascade;
+drop table if exists experiment cascade;
+drop table if exists measurement cascade;
+drop table if exists analysis_method cascade;
+drop table if exists analysis cascade;
+drop table if exists analysis_removed_measurements cascade;
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -1,0 +1,17 @@
+version: '3.7'
+services:
+  postgres:
+    image: "postgres"
+    container_name: postgres
+    ports:
+      - "5432:5432"
+    environment:
+      POSTGRES_USER: ednauser
+      POSTGRES_PASSWORD: ednapass
+      POSTGRES_DB: ednadb
+    volumes:
+      - ./postgres:/var/lib/postgresql/data/
+    networks:
+      - edna-network
+networks:
+  edna-network: { }


pva701:
This will be updated after the 11 Feb meeting

pva701:
should we have a table from compound? It doesn't seem we can get any information about it except for id/name

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,100 @@
+create table project
+(
+    project_id  bigserial     not null primary key,
+    name        varchar(256)  not null unique,
```
serokellcao: AFAIR, in postgresql varchar(N) is just a short-hand for Text, but with 256 characters allocated for display?
If so, it's OK, but otherwise, I'd much rather use Text.

PS I looked at the documentation and it seems like you won't be able to insert anything bigger than 256 symbols into the name. I can imagine a full drug name that's more than 256 symbols. I think that we should default to permissive Text.

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,100 @@
+create table project
```
serokellcao, line 1: Projects form a hierarchy, so you need to track a NULLable parent.

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,100 @@
...
+(
+    project_id  bigserial     not null primary key,
+    name        varchar(256)  not null unique,
+    description varchar(1000) null
+);
+
+create table test
+(
+    test_id bigserial    not null primary key,
+    name    varchar(256) not null unique
```
serokellcao: Just make `subtest` recursive. (See comment below, and see comment about "Projects forming a hierarchy").

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,100 @@
...
+);
+
+create table subtest
```
serokellcao: Why not let tests be recursive (just as projects). So I would request this to be merged up with `test` and renamed to `test`.

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,100 @@
...
+        foreign key (test_id) references test (test_id)
+            on update cascade
+            on delete cascade
+);
+
+-- Corresponds to an EXCEL file
+create table experiment
+(
+    experiment_id bigserial                not null primary key,
+    project_id    bigint                   not null,
```
serokellcao: There is no certainty in regards to how experiments contribute to projects.

But what is certain is that the same microplate can host compounds from different projects that use the same testing method.

So `project_id` is not a property of `experiment`, but rather an `analysis`!

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,100 @@
...
+            on delete no action
+);
+
+create table measurement
+(
+    measurement_id bigserial        not null primary key,
+    experiment_id  bigint           not null,
+    -- Either compound id or name
+    compound_id    varchar(256)     not null,
+    concentration  double precision not null,
```
serokellcao, line 48: I think it would be good to also track repetitions of measurements here.
For example, as our customer said during the meeting, normally the third worksheet will have form

```
Cmp1 Cmp1 Cmp1 Cmp1 Cmp1 ...
Cmp1 Cmp1 Cmp1 Cmp1 Cmp1 ...
Cmp1 Cmp1 Cmp1 Cmp1 Cmp1 ...
Cmp2 Cmp2 Cmp2 Cmp2 Cmp2 ...
Cmp2 Cmp2 Cmp2 Cmp2 Cmp2 ...
Cmp2 Cmp2 Cmp2 Cmp2 Cmp2 ...
...
CmpN CmpN CmpN CmpN CmpN ...
CmpN CmpN CmpN CmpN CmpN ...
CmpN CmpN CmpN CmpN CmpN ...
Ctrl Ctrl Ctrl Ctrl Ctrl ...
0    0    0    0    0    ...
```

Running three repetitions of each compound and some controls at the end of the plate.

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,100 @@
...
+);
+
+create table analysis_outcome
+(
+    outcome_id    bigserial          not null primary key,
+    analysis_id   bigint             not null,
+    experiment_id bigint             not null,
+    compound_id   bigint             not null,
+    -- Maybe JSON better?
+    result        double precision[] not null,
```
serokellcao: Well, result is an interesting beast. TL;DR: I agree that we should track IC50 result in `analysis_outcome` as listed here, but the reason why is non-trivial.

First of all, let's talk about the actual result of conducting of an experiment as far as our knowledge about effectiveness of drugs is concerned.

To begin with, there is no "absolute correct IC50". To determine "absolute IC50" we would need to admit drugs of various concentrations to every single biological unit that has the biological function the drug targets and conduct precise measurements. IC50 itself is a statistical / probabilistic value. By anlysing the data we have, however, we can do analyses that will make us be increasingly sure that IC50 value for a drug we have is predicting what will happen in real world after we admit that drug.

So... when we run an experiment, we already have some IC50 analyses done. Depending on how good we are at choosing only good data for said analyses, we may or may not have a pool of high-quality data and we have some mean IC50 and a value for some deviation (scientists propose standard deviation, but we need to be sure that the distribution of concentration readouts is normal, perhaps in logarithmed fashinon) in the data set. After the experiment is conducted, we add more (hopefully high-quality) data to the pool, making mean IC50 more precise. So the result of our analysis is actually updated tuple of (mean, deviation) for the *compound*.

If we insist on tracking results of a particular analysis, we must to do the following:

  1. calculate IC50 based on the points submitted for analysis
  2. determine where in the distribution lies this calculated value
  
Sadly, point (2) moves over time and it would be a hassle to update it every time a compound gets new high-quality data imported into the database, so *it's OK to just track IC50* of the analysus, but it's *not why we're storing `analysis_outcome`s or analyses in general*.

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,100 @@
...
+-- We can remove this table if we don't need anything apart from IC50
+create table analysis
+(
+    analysis_id bigserial     not null primary key,
+    description varchar(1000) null,
+    -- Analysis-specific parameters
+    parameters  json          null
+);
+
+create table analysis_outcome
```
serokellcao: I wrote a short essay about "what results" are below. But now let's talk a little bit about capturing re-analysis approach!

 1. Not every measurement of each experiment will be accounted for in every analysis that is related to this experiment.
 2. Measurements used should be tracked explicitly.
 3. Relation between analyses and experiments is many-to-one

serokellcao (changes requested):
I request some changes (as seen in inline comments). Hit me up on slack (or here) for discussion.

backend/sql/initial_schema.sql
pva701: where did you get this piece of knowledge? I was under impression that the structure is more or less plain (two level of hierarchy: project and subfolder). If you want just to make as general as possible, I wouldn't do so because it complicates schema and I don't see benefits for MVP. @gromakovsky what do you think?

In general recursive schema are used in case of tree hierarchy as you said, like folder structure but we don't seem to need this

backend/sql/initial_schema.sql
pva701: once again, why? if with project it could make sense (in future, but still unrealistic to me) but tests are really domain specific and they told as explicitly that there is a test and version of test inside (which refer to their confluence). how we should handle case when there is test inside test inside test in frontend? how it's aligned with their domain? Where confluence link should refer to in this case?

I am strongly against doing some general things if they are not required in the domain because this generality causes a lot of corner cases which if happen should be somehow properly handled on both frontend and backend. @gromakovsky what do you think?

backend/sql/initial_schema.sql
pva701: I don't think so, yes relation between plate and subproject is not clear to me as well but from what I heard on the meeting: there might be some quirks with plate correspondence to subfolder but general case is that experiment corresponds to subfolder. We can re-listen to a meeting recording which @gromakovsky made.

>So project_id is not a property of experiment, but rather an analysis!

what? it sounds completely untruth to me. I mean it could be true that in one project used only specific analysis but there is no any explicit relation between them. 
I bet we can listen meeting recording/ask @gromakovsky understanding and we won't find proofs for this statement.

pva701: Analysis in first version is just IC50, so how it's related to project at all? it's just a default thing which is used for any compound in any project as I understand.

backend/sql/initial_schema.sql
pva701: where is repetition in this schema? I mean we have this three sheets which form triple (compound, concentration signal), repetition will be in case of same compound and concentration, it's not explicitly given to us. well, we can check that there are same pairs of compound and concentration but signal might be a little bit different for them, what repetition will be equal to in this case? ok, we can have completely equal triples and but it seems not so possible to me.

I am not sure I understand what controls are maybe it's the source of my confusion.

backend/sql/initial_schema.sql
pva701: I think our main focus for now is to make proper support of storage/hierarcy so this talks about results seems to be premature to me. I mean I don't know how we will re-analyze, whether it will be stored in the db or it's just they removed some points, see IC50 in this case and just disregard it. so I would leave this discussion for later stages when we focus on analysis and all these features about points removal/randomisation.

pva701: so analysis is really tentative in this db schema I left it just in some tentative form. once again: the simplest first version of analysis they want is IC50 value, as far I understood.

backend/sql/initial_schema.sql
serokellcao: @pva701 what's the complication to track parent ID?

I can imagine that a hypothesis will have several attempts to check it as several sub-folders of sub-folder.

backend/sql/initial_schema.sql
serokellcao: @pva701 because SQL migrations are a pain in the ass, significantly more of a pain in the ass than adding one extra field.

Also, I hope that subtests are a mistake and we'll convince them to use a tag (trait) system for test types. I would go as far as to do the following (if you want easy migrations):

*Just use 1-deep tables* and then, when needed, just *add* another field called `parent_id` which will give you a *free* migration!

backend/sql/initial_schema.sql
serokellcao: > We can re-listen to a meeting recording 

Please do, because the person who knows more said "compound ends up in a subproject".

> so how it's related to project at all? it's just a default thing which is used for any compound in any project as I understand.

 1. An instance of analysis is done over one compound and one compound only. 
 2. The same compound can be in different (sub-)projects.
 3. From (1) and (2) follows, that (sub-)projects do not track `experiment` *unless* you consider the same plate readout to *always* have compounds related to a single (sub-)project. This is where we have difference of opinions, but I'm certain that I'm right.

backend/sql/initial_schema.sql
serokellcao: If you look at the initial data file that they sent to us on notion, you will see that every compound have three repetitions (last segment of data with `1/3 2/3 3/3` and the last (row `K`) is control which just shows for some known drug (and some known concentration of active component) yields some *nominal* signal, which serves as an evidence that the experiment is OK.

In this schema, repetitions are horisontal groups of three lines. Your readouts can be something like

```
923 1456 2021 4993 17989 ...
821 1641 1999 5010 16595 ...
904 1399 1989 4892 17031 ...
...
```

The farther to the right, the less is concentration in this case. I'm not too certain how important it is to call it exactly `repetitions`, but it's a standard name in data science. 

I guess, for MVP you can ignore it completely, but it will cause bad UX and someone will ask to add it down the line. There is no reason to not store repetitions per compound per concentration for easier life of users and operators.

backend/sql/initial_schema.sql
serokellcao: > this talks about results seems to be premature to me. I mean I don't know how we will re-analyze

They told it during the meeting while answering to one of the questions, I quote by memory:

"When the file is uploaded it's analysed automatically and the results of analysis is shown to the operator. The operator then can choose to exclude data points and click re-analyse and/or mark the results as good or bad"

So no, we do know how will reanalysing happen.

backend/sql/initial_schema.sql
gromakovsky: (How) Do we track subfolders?

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,100 @@
...
+    name    varchar(256) not null unique
+);
+
+create table subtest
+(
+    subtest_id      bigserial      not null primary key,
+    test_id         bigint         not null,
+    name            varchar(256)   not null unique,
+    -- Method
+    meth            varchar(10000) null,
```
gromakovsky: Is it useful to save 2 characters here? :thinking: Or is `method` reserved?

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,100 @@
+create table project
+(
+    project_id  bigserial     not null primary key,
```
gromakovsky: Minor, but do we actually need it to be :so-big:? Projects are created manually, so I doubt it will ever be greater than 2147483647.

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,100 @@
...
+(
+    test_id bigserial    not null primary key,
+    name    varchar(256) not null unique
+);
+
+create table subtest
+(
+    subtest_id      bigserial      not null primary key,
+    test_id         bigint         not null,
+    name            varchar(256)   not null unique,
```
gromakovsky: I am sorry, must likely that's a stupid question, but since we know that `name` is unique, why do we need `bigserial` ID, can't we use `name` as the primary key? I mean not only here, but also in `test` and `project`.

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,100 @@
...
+    -- Method
+    meth            varchar(10000) null,
+    confluence_link varchar(1000)  null,
+    constraint belongs_to_test
+        foreign key (test_id) references test (test_id)
+            on update cascade
```
gromakovsky: Do we need `on update cascade` here? AFAIU `test_id` is immutable, so if it's somehow updated, we'd like to get an error.

Regarding `on delete cascade`, AFAIU it means that someone may delete a test and automatically delete all its subtests. Maybe that's what we want, maybe we want to explicitly remove all subtests first, I don't know. Probably that's too minor for now, let's consider it later.

backend/sql/initial_schema.sql
gromakovsky: I didn't understand whether an experiment (i. e. one `.xlsx` file) is always limited to one `project` or not.

backend/sql/initial_schema.sql
gromakovsky: @pva701 if there are triples `(lsd, 100, 500)` and `(lsd, 100, 501)`, we can say that the first one has `repetition = 1` and the second one has `repetition = 2`. If it's a standard name in data science and @serokellcao thinks it can be helpful to users, let's add it, it should be easy, right?

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,100 @@
...
+    concentration  double precision not null,
+    signal         double precision not null,
+
+    constraint belongs_to_experiment
+        foreign key (experiment_id) references experiment (experiment_id)
+            on delete cascade
+            on update cascade
+);
+
+-- We can remove this table if we don't need anything apart from IC50
```
gromakovsky: After we got some new data about the current scope of development and decided to focus on experimental data storage rather than analysis stuff, I removed most of analysis generalization from my document (product specification aka discovery report). I think we can really assume IC50, apparently we won't need this generalization any soon.

backend/sql/initial_schema.sql
pva701: method is reserved

backend/sql/initial_schema.sql
pva701: the point of having bigserial not the size but rather it's autoincremented

backend/sql/initial_schema.sql
pva701: https://stackoverflow.com/questions/15477005/what-are-the-pros-and-cons-for-choosing-a-character-varying-data-type-for-primar

It's a classical debate. We can go any way we prefer. By default I prefer approach with surrogate primary key.

backend/sql/initial_schema.sql
pva701: we can make it `no action` (almost the same as `restrict`)

backend/sql/initial_schema.sql
pva701: As we figured out on the last meeting, there might be different kinds of IC50 as well. So we agreed on having this table

serokellcao:
> 
> 
> should we have a table from compound? It doesn't seem we can get any information about it except for id/name

Yes, we should have a table for compounds.

backend/sql/initial_schema.sql
serokellcao: We don't, but we should keep in mind that they want hierarchy.

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,103 @@
+create table project
+(
+    project_id  bigserial not null primary key,
+    name        text      not null unique,
+    description text      null
+);
+
+create table test_methodology
```
serokellcao, line 8: We can change it to `test_type`? Not sure.

serokellcao: I think we should converge all the nomenclature.

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,103 @@
...
+-- Corresponds to an EXCEL file
+create table experiment
+(
+    experiment_id  bigserial                not null primary key,
+    project_id     bigint                   not null,
+    methodology_id bigint                   not null,
+    -- This is input from the user
+    name           text                     null,
+    meta           text                     null,
+    -- This is fallback name to show if name is not specified
```
serokellcao, line 25: I think we should generate informative names for reexport capabilities.

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,103 @@
...
+
+-- Corresponds to an EXCEL file
+create table experiment
+(
+    experiment_id  bigserial                not null primary key,
+    project_id     bigint                   not null,
+    methodology_id bigint                   not null,
+    -- This is input from the user
+    name           text                     null,
+    meta           text                     null,
```
serokellcao: If there is a place for JSON, this is the place for JSON.

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,103 @@
...
+(
+    analysis_method_id bigserial not null primary key,
+    description        text      null,
+    -- Analysis-specific parameters
+    parameters         json      null
+);
+
+create table analysis
+(
+    analysis_id        bigserial          not null primary key,
```
serokellcao: This is the most important table for the first phase of the project, because we want to get the science right quickly and spend majority of time on figuring out good UX for data systematisation and access.

So let's not try to be overly smart and use a very simple, very standard approach to desigining tables like this.

MEASUREMENT_T_ANALYSIS

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,103 @@
...
+
+create table analysis_method
+(
+    analysis_method_id bigserial not null primary key,
+    description        text      null,
+    -- Analysis-specific parameters
+    parameters         json      null
+);
+
+create table analysis
```
serokellcao, line 65: As I said below, it should be an aggregate / meta table, tracking some measurements.
As customers requested, it should have some important metadata: 

 - Confidence of the lab assistant that this data is good vs bad (@gromakovsky that's what I call "high quality data", sets of measurements that were green-lit by a lab assistant)
 - IC50 value that was obtained by this analysis

Also, we'll have important data *derived*:

 - List of other analyses for the same experiment
 - Within how many deviations is the calculated IC50 value from currently established IC50 value for the given compound (see top level review comment)

serokellcao (changes requested):
Analysis tables are still :bardaq:.

I would like to have it cleaned up as I specified.

Also, please add `compound` table. It's not a big deal to store them locally from the standpoint of data consistency, but will improve our life greatly.

Also, when we have `compound` table, we will be able to calculate rolling average and rolling deviation over "good" (or _high quality_ data) with an asynchronous stored procedure.

pva701:
>Also, please add compound table. It's not a big deal to store them locally from the standpoint of data consistency, but will improve our life greatly.

not at all. what information should be there? when a compound should be added there? if it's added automatically (by us), not manually, how is it supposed to help at all? wrong names of compound is still possible for example

pva701:
>Also, when we have compound table, we will be able to calculate rolling average and rolling deviation over "good" (or high quality data) with an asynchronous stored procedure.

how it even comes into you mind? how one table with only one column which is going to be artificially filled by us whenever we met new compound (or when?) is going to IMPROVE life so much that you will be even write some procedures which wasn't possible to write before.

I want hear from you:
- why it will help so much?
- do you suggest us to control (add/remove) compounds whenever they appear or disappear?
- why this redundancy of data is better then implicitly having them?

We can add `view` which would select all existing `compounds`, at least there won't be stale compounds and they will be updated automatically by postgres whenever we update base table.

backend/sql/initial_schema.sql
pva701: we have `analysis_removed_measurements` table. do you want us to original measurements not removed one?

backend/sql/initial_schema.sql
serokellcao: It's not, but we decided that it is.

serokellcao: (For the purposes of MVP)

backend/sql/initial_schema.sql
serokellcao: We decided that it's unnecessary.

backend/sql/initial_schema.sql
serokellcao: Sorry, you're right.

serokellcao:
> 
> 
> > Also, please add compound table. It's not a big deal to store them locally from the standpoint of data consistency, but will improve our life greatly.
> 
> not at all. what information should be there? when a compound should be added there? if it's added automatically (by us), not manually, how is it supposed to help at all? wrong names of compound is still possible for example

It will be added automatically via import from `MDe`. We simply need this data locally, at least to refer back to `MDe`.

backend/sql/initial_schema.sql
serokellcao: Just call make sure that you don't have there anything except for `analysis_id` and `measurement_id` and you PRIMARY KEY over it.

serokellcao:
> 
> 
> > Also, when we have compound table, we will be able to calculate rolling average and rolling deviation over "good" (or high quality data) with an asynchronous stored procedure.
> 
> how it even comes into you mind? how one table with only one column which is going to be artificially filled by us whenever we met new compound (or when?) is going to IMPROVE life so much that you will be even write some procedures which wasn't possible to write before.
> 
> I want hear from you:
> 
>     * why it will help so much?
> 
>     * do you suggest us to control (add/remove) compounds whenever they appear or disappear?
> 
>     * why this redundancy of data is better then implicitly having them?
> 
> 
> We can add `view` which would select all existing `compounds`, at least there won't be stale compounds and they will be updated automatically by postgres whenever we update base table.

This idea even comes into my head because I don't want to use a cache server. Cache slow-to-update data in SQL is perfectly fine and is a standard practice.

You don't need a stored procedure, actually, you should just track 'last_update' in this table with one value per compound and update with a crontab Python script.

I don't agree with `view` because it's unnecessary load to the server that grows over time linearly at best.

backend/sql/initial_schema.sql
gromakovsky: I understand, but why not just `serial`?

backend/sql/initial_schema.sql
gromakovsky: :pepe_sad:

backend/sql/initial_schema.sql
gromakovsky: Since we are not going to display this data (repetition number), it's probably useless to store it.

gromakovsky:
Let me sum up the discussion about having a `compound` table. There are two potential reasons to have it:
1. This way we can record all compounds that have been added to the system. It makes sense iff there is a way to add a compound separately from uploading a file. So far I assumed that the only way to add a compound to the system is to upload a file referring to this compound. If that's the case, we can get all compounds by looking at `measurement`, so this reason is not valid. @serokellcao thinks we need a way to import data from MDe. I don't remember such a requirement and I don't know if we should do it. We should ask I guess.
2. This way we can store additional data for each compound. I am not 100% sure whether we need it now. In particular I am not 100% sure whether MDe and ChemSoft links are generated automatically or or can be provided by the user. Even if we clarify it and get information that we 100% don't need to store extra data in MVP, we can still consider doing that because probably we'll need this data one day, either metadata or extra data to compute something.

backend/sql/initial_schema.sql
gromakovsky: Okay, there was an update saying that we don't need a hierarchy for now, so we can resolve this conversation and get back to it some time later.

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,103 @@
...
+    description         text          null,
+    confluence_link     varchar(1000) null
+);
+
+-- Corresponds to an EXCEL file
+create table experiment
+(
+    experiment_id  bigserial                not null primary key,
+    project_id     bigint                   not null,
+    methodology_id bigint                   not null,
```
gromakovsky: AFAIU most of `bigint`s can't realistically exceed 2147483647, so we can save some space by changing them to `integer`, can't we?

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,103 @@
...
+    -- This is input from the user
+    name           text                     null,
+    meta           text                     null,
+    -- This is fallback name to show if name is not specified
+    raw_file_name  text                     not null,
+    upload_date    timestamp with time zone not null,
+
+    constraint belongs_to_project
+        foreign key (project_id) references project (project_id)
+            on update cascade
```
gromakovsky: Don't we want `on update no action` like in `has_methodology` because `project_id` is immutable? Also why not `restrict`?

backend/sql/initial_schema.sql
```diff
@@ -0,0 +1,103 @@
...
+);
+
+create table analysis
+(
+    analysis_id        bigserial          not null primary key,
+    analysis_method_id bigint             not null,
+    experiment_id      bigint             not null,
+    compound_id        bigint             not null,
+    -- Maybe JSON better?
+    result             double precision[] not null,
```
gromakovsky: It's a list of numbers, right? What does it denote? In case of IC50 computation we will likely have two numbers: IC50 value and deviation. Is it correct that the list will have 2 numbers in this case? Do we have a list here as a generalization because other ways of analysis may produce different results?

If it's done for generalization, I think JSON is better. If it's done for a different reason, please explain it.

backend/sql/initial_schema.sql
pva701: no action and restrict are +- the same, the are some quirks regarding deferred constraints in `restrict`

backend/sql/initial_schema.sql
pva701: >IC50 value that was obtained by this analysis

it's supposed to be in `result` array, as I am not sure which (and how many) values IC50 produces (I think at least two: IC50 and deviation or something like this)

backend/sql/initial_schema.sql
pva701: we can

backend/sql/initial_schema.sql
pva701: >Do we have a list here as a generalization because other ways of analysis may produce different results?

yes and no. I wasn't sure that IC50 produces only 2 values, so I will change it to JSON

backend/sql/initial_schema.sql
gromakovsky: We can have just two numbers there, because that's what we are going to compute in the MVP, or arbitrary JSON, because other analysis methods may produce something more complex and JSON looks like a nice way of abstraction. I am fine with both options.

gromakovsky: > Confidence of the lab assistant that this data is good vs bad 

AFAIU we agreed to call it `is_suspicious` flag.

backend/sql/initial_schema.sql
gromakovsky: Ok, whatever then.

backend/sql/initial_schema.sql
gromakovsky: @pva701 I propose to add a comment saying that we'll need a hierarchy some time later, so that we keep it in mind.

backend/sql/initial_schema.sql
gromakovsky: We have updated the glossary to say "test_methodology" as this table is named here.

gromakovsky (approved):

serokellcao (approved):

PR #5, opened 2021-02-17
Haskell checks on ci

sashasashasasha151:
## Related issue(s)

https://issues.serokell.io/issue/EDNA-11

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,4 +1,16 @@
 steps:
   - label: check trailing whitespace
     commands:
-    - .buildkite/check-trailing-whitespace.sh
+      - .buildkite/check-trailing-whitespace.sh
+  - label: backend server
+    commands:
+      - nix-build -A packages.x86_64-linux.backend-server
+  - label: backend lib
+    commands:
+      - nix-build -A packages.x86_64-linux.backend-lib
+  - label: backend test
+    commands:
+      - nix-build -A checks.x86_64-linux.backend-test
+  - label: backend hlint
+    commands:
+      - nix-shell --run 'hlint backend -j'
--- a/backend/default.nix
+++ b/backend/default.nix
@@ -1,0 +1,15 @@
+{ haskell-nix }:
+let
+  project = haskell-nix.stackProject {
+    src = haskell-nix.haskellLib.cleanGit {
+      src = haskell-nix.cleanSourceHaskell { src = ./.; };
+    };
+  };
+
+  library = project.edna.components.library;
+  server = project.edna.components.exes.edna-server;
+  test = project.edna.checks.edna-test;
+
+in {
+  inherit server library test;
+}
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,162 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1610051610,
+        "narHash": "sha256-U9rPz/usA1/Aohhk7Cmc2gBrEEKRzcW4nwPWMPwja4Y=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "3982c9903e93927c2164caa727cd3f6a0e6d14cc",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "hackage": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1613956326,
+        "narHash": "sha256-SNctpqflg0NKecvS8FcPGQDdPnr3cDzMRLZq+MwKWp0=",
+        "owner": "input-output-hk",
+        "repo": "hackage.nix",
+        "rev": "0c7c981d0512a8759d5381cd8b17ea00e289517f",
+        "type": "github"
+      },
+      "original": {
+        "owner": "input-output-hk",
+        "repo": "hackage.nix",
+        "type": "github"
+      }
+    },
+    "haskell-nix": {
+      "inputs": {
+        "nixpkgs": "nixpkgs",
+        "nixpkgs-2003": "nixpkgs-2003",
+        "nixpkgs-2009": "nixpkgs-2009",
+        "nixpkgs-unstable": "nixpkgs-unstable"
+      },
+      "locked": {
+        "lastModified": 1613996756,
+        "narHash": "sha256-J6JbKCqzpJhlHT3+44vjcvkoX1qJSzOzM31xpS38ysY=",
+        "owner": "input-output-hk",
+        "repo": "haskell.nix",
+        "rev": "c4e5aea19b2436670524a230a0e5e959c8923fd1",
+        "type": "github"
+      },
+      "original": {
+        "owner": "input-output-hk",
+        "repo": "haskell.nix",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1608007629,
+        "narHash": "sha256-lipVFC/a2pzzA5X2ULj64je+fz1JIp2XRrB5qyoizpQ=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "f02bf8ffb9a5ec5e8f6f66f1e5544fd2aa1a0693",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "f02bf8ffb9a5ec5e8f6f66f1e5544fd2aa1a0693",
+        "type": "github"
+      }
+    },
+    "nixpkgs-2003": {
+      "locked": {
+        "lastModified": 1607708579,
+        "narHash": "sha256-QyADEDydJJPa8n3xawnA82IJAcZHNNm6Pp5DU7exMr4=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "7f73e46625f508a793700f5110b86f1a53341d6e",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "7f73e46625f508a793700f5110b86f1a53341d6e",
+        "type": "github"
+      }
+    },
+    "nixpkgs-2009": {
+      "locked": {
+        "lastModified": 1608007629,
+        "narHash": "sha256-lipVFC/a2pzzA5X2ULj64je+fz1JIp2XRrB5qyoizpQ=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "f02bf8ffb9a5ec5e8f6f66f1e5544fd2aa1a0693",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "f02bf8ffb9a5ec5e8f6f66f1e5544fd2aa1a0693",
+        "type": "github"
+      }
+    },
+    "nixpkgs-unstable": {
+      "locked": {
+        "lastModified": 1612284693,
+        "narHash": "sha256-efzJNF1jvjK3BMl0gZ0ZaUWcFMv0nLb9AHN/++5+u0U=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "410bbd828cdc6156aecd5bc91772ad3a6b1099c7",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "410bbd828cdc6156aecd5bc91772ad3a6b1099c7",
+        "type": "github"
+      }
+    },
+    "nixpkgs_2": {
+      "locked": {
+        "lastModified": 1613641638,
+        "narHash": "sha256-dZM2usljlkGNU1krgBXRsQ2tHWZExFPnvecullGe504=",
+        "owner": "serokell",
+        "repo": "nixpkgs",
+        "rev": "8a1d9e6a3596d340636eccbbc2cdf7429d4af632",
+        "type": "github"
+      },
+      "original": {
+        "owner": "serokell",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "flake-utils": "flake-utils",
+        "hackage": "hackage",
+        "haskell-nix": "haskell-nix",
+        "nixpkgs": "nixpkgs_2",
+        "stackage": "stackage"
+      }
+    },
+    "stackage": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1613437366,
+        "narHash": "sha256-EIMjXbwx41TNR/dWEWJrKbO61h17tzLDNwTheTXrHJU=",
+        "owner": "input-output-hk",
+        "repo": "stackage.nix",
+        "rev": "aba6cc7066d2f5d1179284d708e11751f1c1be78",
+        "type": "github"
+      },
+      "original": {
+        "owner": "input-output-hk",
+        "repo": "stackage.nix",
+        "type": "github"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,37 @@
+{
+  inputs = {
+    nixpkgs.url = "github:serokell/nixpkgs";
+    haskell-nix.url = "github:input-output-hk/haskell.nix";
+    flake-utils.url = "github:numtide/flake-utils";
+
+    hackage = {
+      url = "github:input-output-hk/hackage.nix";
+      flake = false;
+    };
+
+    stackage = {
+      url = "github:input-output-hk/stackage.nix";
+      flake = false;
+    };
+  };
+
+  outputs = { self, nixpkgs, haskell-nix, hackage, stackage, flake-utils }:
+    flake-utils.lib.eachSystem [ "x86_64-linux" ] (system:
+      let
+        pkgs = nixpkgs.legacyPackages.${system}.extend haskell-nix.overlay;
+        backend = pkgs.callPackage ./backend { };
+      in {
+        defaultPackage = self.packages.${system}.backend-server;
+        packages = {
+          backend-lib = backend.library;
+          backend-server = backend.server;
+        };
+
+        checks.backend-test = backend.test;
+
+        devShell = pkgs.mkShell {
+          inputsFrom = [ self.packages.${system}.backend-lib ];
+          buildInputs = with pkgs.haskellPackages; [ cabal-install hpack hlint ];
+        };
+      });
+}
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,4 @@
+(import (fetchTarball
+  "https://github.com/edolstra/flake-compat/archive/master.tar.gz") {
+    src = builtins.fetchGit ./.;
+  }).defaultNix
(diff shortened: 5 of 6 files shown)


backend/default.nix
```diff
@@ -0,0 +1,27 @@
+{ haskell-nix, runCommand, lib }:
+let
+  project = haskell-nix.stackProject {
+    src = haskell-nix.haskellLib.cleanGit {
+      src = haskell-nix.cleanSourceHaskell { src = ./.; };
+    };
+    modules = [
+      {
+        doHaddock = false;
```
gromakovsky: I think we want to do Haddock btw.

flake.lock
```diff
@@ -0,0 +1,291 @@
...
+        "rev": "91e934fd9e1e06044f75a3e3ad998813c2a90fb5",
+        "type": "github"
+      },
+      "original": {
+        "owner": "serokell",
+        "repo": "deploy-rs",
+        "type": "github"
+      }
+    },
+    "deploy-rs": {
```
gromakovsky: I know it's not ready for review yet, but keep in mind that many things here are unused (AFAIU) and should be removed, e. g. I doubt we use `deploy-rs`.

.buildkite/pipeline.yml
```diff
@@ -1,4 +1,7 @@
 steps:
-  - label: check trailing whitespace
+#  - label: check trailing whitespace
+#    commands:
+#      - .buildkite/check-trailing-whitespace.sh
```
gromakovsky: Did it stop working?

gromakovsky:
We've decided not to do the check for cabal files in this PR (unless we find a solution quickly).

.buildkite/pipeline.yml
```diff
@@ -1,4 +1,16 @@
 steps:
   - label: check trailing whitespace
     commands:
-    - .buildkite/check-trailing-whitespace.sh
+      - .buildkite/check-trailing-whitespace.sh
+  - label: backend build
```
gromakovsky: What is `backend build` exactly? We have `backend lib`, `backend test` and `backend build`. Is `backend build` supposed to build the library, test, everything or something else? I think there is some redundancy here.

backend/.#default.nix
```diff
@@ -0,0 +1 @@
+balsoft@AMD-Workstation.15930:1613204272
```
gromakovsky: Don't forget to remove it.

backend/default.nix
```diff
@@ -0,0 +1,44 @@
+{ haskell-nix, runCommand, lib, hlint }:
+let
+  ignoreFilter = path: type:
```
gromakovsky: Do you understand why we need this `ignoreFilter`?

backend/default.nix
```diff
@@ -0,0 +1,44 @@
...
+    in
+      !(
+        baseName == ".gitignore" ||
+        hasPrefix "resources" relPath ||
+        hasSuffix ".md" baseName ||
+        hasSuffix ".nix" baseName
+      );
+
+  # Set a constant name for the src closure
+  source = let
```
gromakovsky: AFAIU this `source` thing and the above `ignoreFilter` are only used to run `hlint`, I see it used only in `hlint-html` (and btw, does `${source}` there refer to this `source` or something else?). Is it indeed needed only for `hlint` or is it implicitly used somewhere else? I just Ctrl-F'ed `source`.

I have an alternative solution: just call `hlint` from `pipeline.yaml` using `nix run`. Pros:
1. No need to have all this extra code in this file (`source`, `ignoreFilter`). Though maybe I am wrong and it's needed for something else as well.
2. Less steps involved, easier to understand what's going on, easier to find what arguments are passed and where to change them.

AFAIU there is one problem: we don't know how to make `nix run` work with flakes. If that's the case and we don't figure it out quickly, I am okay with changing it later.

backend/default.nix
```diff
@@ -0,0 +1,44 @@
...
+          doHaddock = true;
+          package.ghcOptions = "-Werror";
+        };
+      }
+    ];
+  };
+  library = project.edna.components.library;
+  server = project.edna.components.exes.edna-server;
+  test = project.edna.checks.edna-test;
+  hlint-html = runCommand "hlint.html" {} "${hlint}/bin/hlint ${source} --no-exit-code --report=$out -j";
```
gromakovsky: 1. What is `hlint.html`? Do you understand?
2. Why do we pass `--no-exit-code`? This way it seems very unlikely to notice when it reports anything. If we don't want CI to fail when `hlint` fails, we can mark the corresponding step as optional in the CI config. Just set `soft_fail: true`. I think it's better because this way it will be easier to notice that `hlint` reported something.

gromakovsky (changes requested):

pipeline.nix
```diff
@@ -0,0 +1 @@
+(import ./.).pipelineFile
```
gromakovsky: Is this file needed?

flake.nix
```diff
@@ -0,0 +1,51 @@
...
+
+      ciSystems = [ "x86_64-linux" ];
+
+      checks = builtins.mapAttrs (system: _:
+        with packagesFor system; {
+          backend-test = backend.test;
+          backend-hlint = backend.hlint;
+        }) nixpkgs.legacyPackages;
+
+      pipelineFile = common-infra.mkPipelineFile self;
```
gromakovsky: And do we need this?

flake.nix
```diff
@@ -0,0 +1,51 @@
...
+
+      devShell = builtins.mapAttrs (system: packages:
+        let pkgs = nixpkgs.legacyPackages.${system};
+        in pkgs.mkShell {
+          inputsFrom = [ packages.backend-lib ];
+          buildInputs = with pkgs.haskellPackages; [ cabal-install hpack ];
```
gromakovsky: I don't use dev shell and I am not sure anybody needs it, but if we decide to keep it, probably we don't need `hpack` here because we don't use it (no `package.yaml`).

flake.nix
```diff
@@ -0,0 +1,51 @@
...
+      url = "github:input-output-hk/stackage.nix";
+      flake = false;
+    };
+    common-infra.url = "github:serokell/common-infra";
+  };
+
+  outputs = { self, nixpkgs, haskell-nix, hackage, stackage, common-infra }:
+    let
+      packagesFor = system: rec {
+        pkgs = nixpkgs.legacyPackages.${system}.extend
```
gromakovsky: What is the semantic of this `pkgs`?

flake.nix
```diff
@@ -0,0 +1,51 @@
...
+            haskell-nix.overlay
+          ]);
+        backend = pkgs.callPackage ./backend { };
+      };
+    in {
+      packages = builtins.mapAttrs (system: _:
+        with packagesFor system; {
+          backend-lib = backend.library;
+          backend = backend.server // {
+            meta.modulePath = [ "services" "edna" "backend" ];
```
gromakovsky: What is `meta.modulePath`?

flake.nix
```diff
@@ -0,0 +1,51 @@
...
+          };
+        }) nixpkgs.legacyPackages;
+
+      ciSystems = [ "x86_64-linux" ];
```
gromakovsky: Btw, it seems very unlikely that we will have more than one `ciSystem` in any foreseeable future, so this file can be simplified, we don't need to iterate over systems and pass them to functions, we can have a single constant `ciSystem`.

flake.lock
```diff
@@ -0,0 +1,185 @@
...
+        "rev": "ba87ffb84bc4158ada14908bddee9df2c81ac31b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "input-output-hk",
+        "repo": "haskell.nix",
+        "type": "github"
+      }
+    },
+    "naersk": {
```
gromakovsky: https://github.com/nmattia/naersk says "Nix support for building cargo crates". Do we need it?

gromakovsky:
@mkaito I have a question to you.
In many repos we often do things like `nix run -f ci.nix pkgs.reuse -c reuse lint`. By passing `pkgs.reuse` here we ensure that `reuse` is available in `$PATH`. To make it work we need to define `pkgs` in `ci.nix`, so that it has `reuse` attribute. In `ci.nix` we have something like this:
```
  sources = import ./nix/sources.nix;
  haskellNix = import sources."haskell.nix" {
    sourcesOverride = { hackage = sources."hackage.nix"; stackage = sources."stackage.nix"; };
  };
  pkgs = import sources.nixpkgs haskellNix.nixpkgsArgs;
```

So AFAIU `pkgs` is `nixpkgs` and its revision is taken from `nix/sources.json`. That works well when we use `niv`.
I like this `nix run -f ci.nix INSTALLABLE -c bla-bla-bla` feature and I think we want to use it in this repo. But it's not clear what's the best way to do the same with flakes instead of `niv`.
First of all, since we have `default.nix` instead of `ci.nix`, we don't need to pass `-f ci.nix`, that's clear to me.
But it's not clear how to access nixpkgs like we have `pkgs` in the approach with `niv`. We have `defaultNix` in `default.nix`, so AFAIU its contents is fully defined by `flake.nix`. Since we have `nixpkgs` in flake's `outputs`, can we just write `nix run nixpkgs.reuse`? Or should we do it differently?

P. S. Example with `niv`: https://github.com/tqtezos/baseDAO

gromakovsky:
> Since we have `nixpkgs` in flake's `outputs`

Ah, oh, sorry, it's not in `outputs`, it's in **arguments** of `outputs`, so `outputs` expects `nixpkgs` to be passed to it.

I've just tested my hypotheses. First I tried to just do `nix run nixpkgs.hlint -c hlint`, see [here](https://buildkite.com/serokell/edna/builds/86#fe1be568-d905-49d6-bee1-97dcbb59bad9). It said:
> attribute 'nixpkgs' in selection path 'nixpkgs.hlint' not found

Then I tried to also add `nixpkgs = nixpkgs;` (not sure if it makes sense), but got the same message [here](https://buildkite.com/serokell/edna/builds/87#ddec1278-729e-4b0f-b568-60fb4b577aed). You can also see commits I made, they are available from buildkite.

I don't want to go further with experiments, will wait for Chris to respond.

flake.nix
mkaito: `devShell` is the Flake equivalent of `shell.nix` -- and yes, I use it all the time.

flake.lock
mkaito: It's used by `deploy-rs`. `flake.lock` contains all transitive dependencies for all inputs in the flake. `deploy-rs` is a transitive dependency of `common-infra` which is an input to this flake.

gromakovsky: AFAIU we don't use `deploy-rs` either, but I see that it's a dependency of `common-infra`, so ok. It's a bit poor granularity (that we need to mention so many packages just for a small piece of `common-infra`), but I guess we will end up using `deploy-rs` anyway, so it's not a problem.

mkaito:
The command `nix run` changed semantics in 3.0. In 2.4 it was a replacement for `nix-shell -p blah --run whatever`, but in 3.0 it does something else entirely.

It relies on a flake concept called "apps", where you define executables in a flake output that can be run with `nix run` without installing. For example, you can `nix run github:serokell/deploy-rs` to run that program directly from git master without installing it.

What you're looking for is now called `nix shell`.

This will do what you want.

```sh
$ nix shell nixpkgs#hlint -c hlint
```

Note the `nixpkgs#hlint` reference. This puts the package `hlint` from the flake `nixpkgs` in `$PATH` and then calls command `hlint`. The `nixpkgs` flake is defined in the flake registry as `github:nixos/nixpkgs`.

Generally speaking, all our inputs should be declared in the flake, and everything should be run through the flake. The `mkPipelineFile` function will build a sepate CI step for each package, check, and deployment attribute in the flake, so you should just plug things into `checks` as necessary. I'm not sure it supports custom extra steps, but that can probably be added if necessary.

The files `default.nix` and `shell.nix` contain compatibility shims to allow Nix 2.4 to read flake outputs, so you can `nix-build` and `nix-shell` without installing 3.0.

gromakovsky:
> The `mkPipelineFile` function will build a separate CI step for each package

What do you mean by "build a separate CI step"? Does it generate `.pipeline.yml`? Is it supposed to be appended to the main `.pipeline.yml`? Or replace it? Or what? How does it generate steps, how does it know what options to specify for steps, like `soft_fail`, `if`, etc.?

gromakovsky:
> This will do what you want.

Ok, let me test it in my branch.

mkaito:
> Does it generate .pipeline.yml?

Yes.

> How does it generate steps

By reading the `packages` and `checks` outputs in the flake and generating one CI step for each. Here's what it currently generates (`nix build .#pipelineFile && cat result | jq`):

```json
{
  "steps": [
    {
      "agents": [],
      "command": "nix-build -A packages.x86_64-linux.backend-lib",
      "label": "Build backend-lib"
    },
    {
      "agents": [],
      "command": "nix-build -A packages.x86_64-linux.backend-server",
      "label": "Build backend-server"
    },
    {
      "agents": [],
      "command": "nix-build --no-out-link -A checks.x86_64-linux.backend-hlint",
      "label": "backend-hlint"
    },
    {
      "agents": [],
      "command": "nix-build --no-out-link -A checks.x86_64-linux.backend-test",
      "label": "backend-test"
    },
    "wait"
  ]
}


```

> how does it know what options to specify for steps

It doesn't. It just runs what you tell it to run. For example, `checks.backend-hlint` is defined in `backend/default.nix` as such:

```nix
{
  hlint-html = runCommand "hlint.html" {} "${hlint}/bin/hlint ${source} --no-exit-code --report=$out -j";
}
```

flake.lock
mkaito: `common-infra` has 2 nix functions: `mkNode` to define deployment nodes, and `mkPipelineFile` that generates CI steps which uses the former to generate deployment steps. So we're technically using half of `common-infra` :stuck_out_tongue: 

Packages mentioned in `flake.lock` don't necessarily end up in any of the closures, and since Nix is lazy, they don't even get evaluated unless you actually use them somewhere. The lockfile is eager and will contain all transitive locks regardless of whether they are used. It's just extra noise, that's all.

backend/default.nix
mkaito: 1. It's the name of the output file generated in the closure for the `hlint-html` derivation. It's an arbitrary name. The important part is the closure itself and its hash, which depends on the inputs defined in `source`.

2. No idea. This looks like what Yegor wrote years ago for Disciplina. I suspect it has just been cargo culted ever since. Feel free to adjust the invocation of `hlint` as you see fit. Just make sure you put something in `$out` somehow.

.buildkite/pipeline.yml
mkaito: It builds `backend.server`, according to the way things are declared now. This file won't be used either way, as we'll use the generate pipeline steps.

backend/default.nix
mkaito: Cargo culted from something I wrote in 2017. It's not necessary, and I'm removing it.

flake.nix
mkaito: We do, actually.

flake.nix
mkaito: Flakes define derivations for all known systems. In the context of a flake, `nixpkgs` isn't a set of packages. It's a set of sets of packages. So you need to use the set for the current (for some definition of "current") system.

flake.nix
mkaito: No idea. Removed it.

flake.nix
mkaito: This isn't used anywhere. `mkPipelineFile` only generates steps for `x86_64-linux`.

gromakovsky:
So does it replace `.pipeline.yml`? Can we refrain from doing that please and just write `.pipeline.yml` manually? I don't like it for many reasons:
1. If I just want to do something simple on CI (e. g. call some `.sh` script), I need to go to some nix file and do it there. That's overhead and entry barrier.
2. Developers generally don't like dealing with nix, so each use of nix should be justified. Here we just move some configuration from yaml to nix files, so I don't see a big benefit.
3. Our CI approach changes every year, some details change every month. This `mkPipelineFile` thing is new (AFAIU) and not battle-tested (in many repos for a long time). I don't want to use such things in this project where we just need to quickly do what we know how to do and don't have much space for experiments. I am tired in this never-ending changes in the way we handle CI.
4. If we want to specify custom `agents` or `if` or set `soft_fail` or any other option, is there any way to do it? AFAIU you are saying there is none.

> For example, `checks.backend-hlint` is defined in `backend/default.nix` as such:

Yes, and I don't like that, don't understand why it's done this way and proposed to change that. I like
```
  - label: hlint
    commands:
      - nix shell nixpkgs#hlint -c hlint
```

much more.

mkaito:
I've cleaned up some things and removed pipeline generation.

gromakovsky:
@mkaito thank you, that's exactly what I wanted to see. LGTM now. One last question I wanted to ask: I see the things in flake's output are called `packages` and `checks`. Is it flake's conversion or our own? I was thinking that these names are treated specially by `mkPipelineFile`: it passes `--no-out-link ` for `checks` and doesn't do it for `packages`. Is there anything else where these names matter?

If I rename `packages` to `foo` and `checks` to `bar` and update `.pipeline.yml` accordingly, will it work the same way? I am ok with current names and don't suggest changing them, just want to know whether there are any other conventions.

gromakovsky (approved):
I'm ok with merging it given that:
1. You fix CI so that it passes (obviously).
2. Commits are cleaned up.
3. @sashasashasasha151 reviews latest changes and confirms he's satisfied with them.

mkaito:
> I see the things in flake's output are called packages and checks. Is it flake's conversion or our own?

There are outputs that are treated specially. `packages` are build outputs, while `nix flake check` (need nix 3.0) will eval everything referenced in the flake, build all packages and also all checks. There's other meaningful outputs, such as `nixosModules` and `overlay`. Flakes present a standardized interface to modular repositories of NixOS components. See doc links further down.

> it passes --no-out-link for checks and doesn't do it for packages.

The `result` symlink act as a GC root. As long as something in the store is referenced by a gc root, it will not get collected. We pin closures for packages but not for tests. The rationale being that packages are used for deployment and we don't want to rebuild them if a GC happens at the wrong time.

> Is there anything else where these names matter?

Yeah, although flakes are sparsely documented. Half of what we know we learned by reading the C++ source code of Nix itself. The best documentation I've found is oddly enough `nix flake --help`, for which you need a very recent version of the Nix client installed.

The de-facto standard documentation is the old and closed [RFC 49](https://github.com/NixOS/rfcs/pull/49). There's a half decent [wiki page](https://nixos.wiki/wiki/Flakes) too, and some [tutorial style blog posts](https://www.tweag.io/blog/2020-05-25-flakes/).

The Nix folks keep pussyfooting about Flakes. Meanwhile, the rest of the Nix world has been using them a long time and are quite happy. They actually just moved the target for flake release from 3.0 to 3.5 for no apparent reason other than more bikeshedding. Never change, Nix. Never change...

> If I rename packages to foo and checks to bar and update .pipeline.yml accordingly, will it work the same way?

Yes they will, because we don't use any flake specific features in the current pipeline. The shim in `default.nix` presents things in the "old" way, and we just build the derivations with the right names and that's it. This is necessary because our servers run stable nix. We do try however to make it as easy as possible to switch to flake-native Nix in the future, hopefully without having to rewrite every single pipeline. All of these compat shims are used now because flakes require unstable nix, and we don't want to run unstable nix daemon on our servers.

It becomes important once you use a Nix 3.0 client and want to use things like `nix flake check`, which would eval everything, build all packages (for all architectures!) and also all checks.

Some of the flakes I've written for various projects keep a pinned master build of Nix in the devShell so that we can run things like flake checks directly without a shim, but we haven't quite standardized on how exactly we want to do these things.

PR #6, opened 2021-02-19
Parse excel file

pva701:
## Description

What has been done:
- [x] old format parsing implemented
- [x] servant machinery defined (including endpoint to upload a file)
- [x] initial config defined
- [x] set up swagger schema 
- [x] simple frontend to show upload results

## Related issue(s)

https://issues.serokell.io/issue/EDNA-12

- Tests
        silently reappearing again.

- Documentation


backend/src/Edna/Web/API.hs
```diff
@@ -0,0 +1,34 @@
+{-# LANGUAGE DataKinds #-}
+
+{-|
+Servant type-level specification for Edna API.
+-}
+module Edna.Web.API
+     ( EdnaAPI
+     , EdnaEndpoints (..)
+     , ednaAPI
+     ) where
```
gromakovsky: Formatting of exports is inconsistent and doesn't match style guide in most modules :cry:

backend/edna.cabal
```diff
@@ -30,11 +30,29 @@ common common-options
...
+      Edna.Web.API
+      Edna.Web.Error
+      Edna.Web.Handlers
+      Edna.Web.Types
+      Edna.Util
   hs-source-dirs:
       src
+  build-depends:
+    aeson,
+    aeson-options,
```
gromakovsky: `aeson-options` is [deprecated](https://github.com/serokell/aeson-options), please let's not use it, let's use [aeson-casing](https://hackage.haskell.org/package/aeson-casing) instead.

backend/app/Main.hs
```diff
@@ -4,7 +4,28 @@ module Main
...
 
 main :: IO ()
-main = print stub
+main = do
+  configPath <- execParser $
+    info (helper <*> configPathParser) $
+    fullDesc <> progDesc "Edna API server."
+
+  Y.decodeFileEither @EdnaConfig configPath >>= \case
+    Left e -> putText (show e)
```
gromakovsky: In this case we will exit with 0 code which is not what we want. Also the error message should probably go to stderr. Why not use `decodeFileThrow` which does the right thing we want here?

backend/src/Edna/Config.hs
```diff
@@ -0,0 +1,32 @@
...
+
+import Data.Aeson.TH (deriveJSON)
+import Data.Aeson.Options (defaultOptions)
+
+import Edna.Util (NetworkAddress)
+
+data ApiConfig = ApiConfig
+  { acListenAddr :: NetworkAddress
+  , acServeDocs :: Bool
+  } deriving stock (Generic, Read, Show)
```
gromakovsky: What do we need `Read` for?

backend/src/Edna/Web/Handlers.hs
```diff
@@ -0,0 +1,90 @@
...
+          _ -> Left UnexpectedCellType
+      )
+
+  signals <-
+    readSheet'
+      "Signal"
+      ( \case
+          CellDouble x -> Right x
+          _ -> Left UnexpectedCellType
+      )
```
gromakovsky: Apparently you always pass the same lambda to `readSheet'`, can you extract it into a named function?

backend/stack.yaml
```diff
@@ -2,3 +2,8 @@ resolver: lts-17.2
 
 packages:
 - .
+
+extra-deps:
+  - aeson-options-0.1.0
+
+allow-newer: true
```
gromakovsky: Is it needed for `aeson-options` (that we shouldn't use because it's deprecated) or something else?

gromakovsky (changes requested):

backend/src/Edna/Web/Handlers.hs
```diff
@@ -0,0 +1,90 @@
...
+ednaHandlers :: EdnaHandlers Handler
+ednaHandlers =
+  genericServerT
+    EdnaEndpoints
+      { eeUploadExperiment = uploadExperiment
+      }
+
+uploadExperiment :: MultipartData Mem -> Handler [ExperimentalMeasurement]
+uploadExperiment multipart = do
+  let file = U.head (files multipart)
```
gromakovsky: Is it guaranteed to be non-empty?

gromakovsky:
> tests for parsing

Note that the format will change, so don't put too much effort into it. I mean, I think it still makes sense to add tests now, so that we have necessary code in place and can simply update test files later. But let's have just one sample file for now.

backend/src/Edna/Web/API.hs
pva701: It was a test to check your attention (to check that you are alive at all, I think ability to being able to notice such things determine your state being alive).

I will reformat to 2 spaces everywhere in the code.

backend/stack.yaml
pva701: yes, I will remove if it's needed for `aeson-casing`

gromakovsky:
> simple frontend to show upload results

Note that it's a separate issue, [EDNA-13](https://issues.serokell.io/issue/EDNA-13).

backend/src/Edna/Web/Error.hs
```diff
@@ -0,0 +1,45 @@
...
+class Exception e => ToServerError e where
+    toServerError :: e -> ServerError
+
+instance Buildable EdnaServerError where
+  build (XlsxParingError err) = "Xlsx parsing error: "+| err |+""
+
+instance Buildable ExperimentParsingError where
+  build UnexpectedCellType = "Expected one type of cell but got another one"
+  build InvalidCell = "Undefined content of the cell"
+  build (WorksheetNotFound w) = "" +| w |+ " worksheet not found"
```
gromakovsky: Is `"" +| ` really needed here?

backend/config.yaml
```diff
@@ -0,0 +1,3 @@
+api:
```
gromakovsky, line 1: What do you think about also having default config so that one can run `edna-server` without thinking about config and still get something meaningful?

backend/src/Edna/Web/API.hs
```diff
@@ -0,0 +1,34 @@
...
+    eeUploadExperiment :: route
+      :- "experiment"
+      :> Summary "Upload an EXCEL file describing one experiment"
+      :> MultipartForm Mem (MultipartData Mem)
+      :> Post '[JSON] [ExperimentalMeasurement]
+  } deriving stock (Generic)
+
+-- | API type specification.
+type EdnaAPI =
+  "api" :> "v1" :> ToServant EdnaEndpoints AsApi
```
sashasashasasha151: What for we need `"api" :> "v1"`?

backend/src/Edna/Web/Handlers.hs
```diff
@@ -0,0 +1,99 @@
+module Edna.Web.Handlers
```
sashasashasasha151, line 1: I propose to think about structure of the app from the very beginning\
I know that it is MVP and we do not need to think about unrelated things but I think that even in MVP it is important\
We had this problem in blend app and I don't want to meet it here when you can't understand what is going on and your handler file has 1000+ lines and you want to rewrite all backend\
So, lets split all features to different parts. For example here we have part that works with files and it looks like a separate thing\
Also, I propose to use MVC model from start\
First layer will be DB, it will just make queries and transform DB types to domain types.\
Second layer is for business logic, it only works with domain types\
Third layer is for API which just works with business logic and construct types for response\
@gromakovsky what do you think about it?\
@pva701 I do not suggest you to think about it in this PR as at least we do not have DB now (but I already see that it is possible to split handler on two modules), but I also want to here your opinion about it

backend/src/Edna/Web/Handlers.hs
```diff
@@ -0,0 +1,90 @@
...
+import Edna.Web.API (EdnaEndpoints(..))
+import Edna.Web.Types (ExperimentalMeasurement(..))
+
+type EdnaHandlers m = ToServant EdnaEndpoints (AsServerT m)
+
+-- | Server handler implementation for Edna API.
+ednaHandlers :: EdnaHandlers Handler
+ednaHandlers =
+  genericServerT
+    EdnaEndpoints
```
sashasashasasha151:
```suggestion
ednaHandlers = genericServerT EdnaEndpoints
```

backend/src/Edna/Web/Handlers.hs
```diff
@@ -0,0 +1,90 @@
...
+      { eeUploadExperiment = uploadExperiment
+      }
+
+uploadExperiment :: MultipartData Mem -> Handler [ExperimentalMeasurement]
+uploadExperiment multipart = do
+  let file = U.head (files multipart)
+  putStrLn $ "Excel file name " ++ show (fdFileName file)
```
sashasashasasha151, line 33: What about adding logs (and also log api calls)?

backend/src/Edna/Web/Handlers.hs
```diff
@@ -0,0 +1,90 @@
...
+      traverse (maybe (Left InvalidCell) unwrap . _cellValue) (_wsCells sheet)
+
+    mergeCells
+      :: Map (Int, Int) Text
+      -> Map (Int, Int) Double
+      -> Map (Int, Int) Double
+      -> [ExperimentalMeasurement]
+    mergeCells compounds concentrations signals =
+      mapMaybe
+        (\(rc, cmp) -> ExperimentalMeasurement cmp <$> M.lookup rc concentrations <*> M.lookup rc signals)
```
sashasashasasha151: Too long line

backend/app/Main.hs
```diff
@@ -4,7 +4,28 @@ module Main
 
 import Universum
 
-import Stub (stub)
+import qualified Data.Yaml as Y
```
sashasashasasha151, line 7: Why not using `Loot`?

sashasashasasha151 (changes requested):

sashasashasasha151:
Also @pva701, why you don't add task name to commits and branch name?

gromakovsky:
https://github.com/serokell/edna/pull/1#issuecomment-776822123

backend/src/Edna/Web/Handlers.hs
gromakovsky: I think this module should have almost no logic and just define `ednaHandlers` with simple definitions. `eeUploadExperiment = bla-bla` where `bla-bla` is something small and uses logic from another module. So in this case I think `parseExperimentXls` should be in a different module, somewhere outside of `Edna.Web`. I think @pva701 just didn't care much about the structure of modules here, which is ok for this PR (as you said).

Overall I agree with this 3-layer approach, but currently there is just no code to apply it :)

backend/src/Edna/Web/Handlers.hs
pva701: what about not to do ALL the initialisation work in completely unrelated task? we can have this un-merged for a while until all this stuff will be initialised.

feel free to create a task to add it.

backend/app/Main.hs
pva701: because it's piece of overcomplicated shit. Do you even know why loot config was created and why do we need to use it at all? (I really doubts)

also I tried to use it with current resolver and there are a lot of compile time errors. feel free to add it yourself :putya:

frontend/.eslintrc.js
```diff
@@ -20,6 +20,9 @@ module.exports = {
     // Sometime we need to specify any type, so it's rather useful
     "@typescript-eslint/no-explicit-any": 0,
 
+    "react/jsx-closing-bracket-location": 0,
+    "react/no-array-index-key": 0,
```
sashasashasasha151: https://stackoverflow.com/questions/46735483/error-do-not-use-array-index-in-keys \
Please do not remove it

frontend/.eslintrc.js
```diff
@@ -20,6 +20,9 @@ module.exports = {
     // Sometime we need to specify any type, so it's rather useful
     "@typescript-eslint/no-explicit-any": 0,
 
+    "react/jsx-closing-bracket-location": 0,
```
sashasashasasha151, line 23: Why?

frontend/src/utils/utils.ts
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line import/prefer-default-export
+export function isDefined<T>(anyVal: T | undefined | null): anyVal is T {
```
sashasashasasha151: Where are you using it?

frontend/src/pages/upload/UploadPage.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <div className="uploadPageContainer">
+      <form
+        className="uploadForm"
+        onSubmit={async (e) => {
+          e.preventDefault();
+          try {
+            if (uploadInputRef.current
+          && uploadInputRef.current.files
+          && uploadInputRef.current.files.length > 0) {
```
sashasashasasha151, line 28: Our config is bad) \
Will take a look later

frontend/src/components/Spinner.scss
```diff
@@ -0,0 +1,61 @@
+// Taken from here https://projects.lukehaas.me/css-loaders/
```
sashasashasasha151, line 1: It is just for example, right?

sashasashasasha151 (changes requested):

frontend/.eslintrc.js
pva701: what you suggest using instead? Keys are a pain in the ass.

frontend/src/utils/utils.ts
pva701: to avoid writing `x != null && x != undefined` whenever I want to check it. 
(don't suggest if (x) because empty strings, lists, zero values will be implicitly converted to false in this case)

what you suggest using instead?

frontend/.eslintrc.js
pva701: because construction
```
<Comp
    attr1=..
    attr2=..
>
  .. code here
</Comp>
```

looks ugly, what is the point of moving > at the separate line? it's not improve readability to me. does it do for you? it just looks like nonsense to me. why should we keep this rule?

frontend/src/components/Spinner.scss
pva701: yes, just random filler to show it's uploading

frontend/src/utils/utils.ts
pva701: oh, yes, nowhere. perhaps worth removing

backend/config.yaml
pva701: well, I think we can use default config if a config file is not found or something.

backend/src/Edna/Web/API.hs
pva701: perhaps, it's redundant indeed

backend/src/Edna/Web/Handlers.hs
pva701: good point, though, I can't separate this uploading logic from the rest of (future) logic (perhaps because there is no logic yet and I can't come up with it). so I more or less agree but don't want to do it in this PR :putya:

backend/src/Edna/Web/Handlers.hs
pva701: btw, it's how Ormolu in idea formatted it :putya: I will fix

backend/src/Edna/Web/Handlers.hs
pva701: like @jagajaga's nick ? :putya:

backend/edna.cabal
pva701: I just wanna notice how Jonn and Arseniy are just having fun together in this PR :putya:

backend/src/Edna/Web/Handlers.hs
gromakovsky: What about just creating a new module and putting `parseExperimentXls` there?

frontend/.eslintrc.js
sashasashasasha151: If it is just a dummy component you can use `index.toString()`\
What index we will use later? Maybe generate indexes on backend or you can also generate it on frontend. The problem is that if we will do some changes in this table or somehow filter/sort rows we will see how it doesn't work)

frontend/.eslintrc.js
sashasashasasha151: I don't think that it is ugly as it helps you to quickly separate `attrs` from `code here`, but if it really important for you I am ok with it

backend/src/Edna/Web/Handlers.hs
gromakovsky: Whatever, we'll do it along with implementing proper parser (now that we have example of inputs).

gromakovsky (approved):
Don't want to delay merging further.

README.md
```diff
@@ -23,6 +23,10 @@ There is nothing to build yet.
 To run a postgres container run `docker-compose up postgres` in `deployment` directory,
 then you can connect to it via `psql` or using any convenient database viewer.
 
+To run a server run `stack exec -- edna-server -c config.yaml`:
+* The server endpoints available at `localhost:9000/api/v1/*`.
```
sashasashasasha151: Remove v1

sashasashasasha151 (approved):

PR #8, opened 2021-03-01
DB schema integration to backend

sashasashasasha151:
## Related issue(s)

https://issues.serokell.io/issue/EDNA-20

- Tests
        silently reappearing again.

- Documentation


backend/config.yaml
```diff
@@ -1,3 +1,8 @@
 api:
   listen-addr: "*:9000"
   serve-docs: true
+
+db:
+  # Syntax: https://www.postgresql.org/docs/9.5/libpq-connect.html#LIBPQ-CONNSTRING
+  conn-string: "host=localhost dbname=edna user=postgres password=12345"
```
pva701: these credentials don't match on in `deployment/docker-compose.yml`, I think it makes sense to make them same (fixing docker ones or these one, it doesn't matter to me)

Also, I am just wondering why you din't use `docker-compose up postgres`, it's so easy to run, no need to fuck with local db (and setting up new db). Nvm.

backend/src/Edna/DB/Connection.hs
```diff
@@ -0,0 +1,45 @@
...
+import Edna.Util (ConnString(..))
+
+-- | Database connection pool. One @Connection@ can not be used simultaneously
+-- by multiple threads, so we need a pool of connections to allow for
+-- simultaneous DB queries.
+newtype ConnPool = ConnPool
+  { unConnPool :: Pool Connection
+  } deriving stock (Show)
+
+withPostgresConn :: EdnaConfig -> IO ConnPool
```
pva701: lol, what's a beauty. Is it a copy-paste?

In any case, I don't think this function makes any sense: you create a pool, then return it from bracket then destroy, you will end up invalid pool (as it was destroyed)

I suspect the name was copy-pasted but semantic changed.

I think it makes sense to add action `ConnPool -> IO a` as a parameter to make it working (at least function which start with `with` usually have such semantic)

pva701: Also, maybe makes sense to make this function working in `RIO` then you don't need to pass config. but it does depend on the use case when you are going to use it

backend/src/Edna/Setup.hs
```diff
@@ -1,28 +1,41 @@
...
 
+setEncoding :: IO ()
+setEncoding = do
+  hSetEncoding stdout utf8
+  hSetEncoding stderr utf8
+
 runEdna :: EdnaConfig -> Edna a -> IO a
 runEdna config action = do
+  setEncoding
+  pool <- withPostgresConn config
```
pva701: I think when/if you make this function to accept an action, this piece of code will change slightly but it will work at least :putya:

backend/src/Edna/DB/Integration.hs
```diff
@@ -0,0 +1,66 @@
...
+
+withConnection :: (Connection -> Edna a) -> Edna a
+withConnection action = do
+  (ConnPool pool) <- (^. edConnectionPool) <$> ask
+  withRunInIO $ \unlift -> withResource pool (unlift . action)
+
+transact :: Edna a -> Edna a
+transact action = withConnection $ \conn -> withRunInIO $ \unlift -> withTransactionSerializable conn (unlift action)
+
+runPg :: Pg a -> Edna a
```
pva701, line 41: I give you 100 if you rename it to `runPog` (stands for POstGres)

backend/sql/initial_schema.sql
```diff
@@ -1,11 +1,11 @@
-create table project
+create table if not exists project
```
pva701, line 1: it seems like one more heritage of cargo cult in our company: I don't think if a table exists while we are trying to create means something good for us and it's worth ignoring this fact (as it will happen with `if not exists`)
Most prolly:
1. we didn't clean up it properly previous time (like there is a forgotten table in drop.sql)
2. this table corresponds to a previous version of sql file (previous version of db schema)

Your considerations to add this `if not exists`? is there some solid reason to do it?

backend/src/Edna/DB/Schema.hs
```diff
@@ -0,0 +1,14 @@
...
+import Universum
+
+import Data.FileEmbed (embedStringFile)
+import Database.Beam.Backend (MonadBeam, runNoReturn)
+import Database.Beam.Postgres (Postgres)
+import Database.Beam.Postgres.Syntax (PgCommandSyntax(..), PgCommandType(..), emit)
+
+ensureSchemaIsSetUp :: MonadBeam Postgres m => m ()
+ensureSchemaIsSetUp = runNoReturn $
+  PgCommandSyntax PgCommandTypeDataUpdate $ emit $(embedStringFile "sql/initial_schema.sql")
```
pva701: on, one more heritage of blind cargo cult. I have no idea what benefits embedding of sql during compile time file gives us. It's not even type-checked (that sql is correct), and any modification (even adding break-line in the end) will cause recompilation the whole project. 
And most annoying thing, that when you commit this file modified, it will cause modification of the file attributes (or something like that) and it will cause recompilation one more time (even if you compiled code with the same file before the commit).

but I don't care a lot because hopefully I am not going to touch this piece of ~garbage~ software. 

let @gromakovsky to decide

backend/src/Edna/Web/Server.hs
```diff
@@ -50,9 +52,9 @@ translateExceptions action =
 -- | Runs the web server which serves Edna API.
 edna :: Edna ()
 edna = do
+  transact (runPg ensureSchemaIsSetUp)
```
pva701: now I see, that `if not exists` is a hack to check that we initialised db before.

If I was working on that, I would add `mode` field to a config file, and if `recreate-db` is passed as mode: drop and create db from scratch (maybe with some dummy data) but without `if not exists` in sql file. 

It seems to me that adding new table to sql file and not adding it to drop sql file, then modifying some fields of the table on the sql, and as consequence, not recreation of the table is possible case during development. It's not a big deal because it will be caught during review, but it can just save some time on figuring out what is going on. (Now `if not exists` will just swallow the fact it already exists).

But I know it's almost impossible to fight some established stupid patterns in the company, I don't mind you leaving it as it's as a tribute to the Blender which was born from Agora, which, in turn, was born from Disciplina. All these guys deserve tribute :putya:.

Jokes aside, I think it's up to @gromakovsky to decide whether it makes sense doing or not

README.md
```diff
@@ -18,6 +18,10 @@ Edna is an open-source tool for data analysis aimed to help researchers with the
 
 There is nothing to build yet.
 
+## Tests
```
pva701: can be squashed with previous commit.

feel free to ignore

pva701 (dismissed):
I think fixing bug in `withPostgresConn` worth doing (if there is a bug indeed).

The rest is up to you and @gromakovsky

backend/default.nix
```diff
@@ -1,9 +1,33 @@
...
       src = haskell-nix.cleanSourceHaskell { src = ./.; };
     };
+    modules = [{
+      doHaddock = true;
+      packages = {
+        edna = {
+          doHaddock = true;
+          ghcOptions = ["-Werror"];
+          components.tests.edna-test = {
+            ghcOptions = ["-Werror"];
```
gromakovsky, line 14: Do we need it both here and 2 lines above?

backend/scripts/test.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+# Runs blnd-tender tests.
```
gromakovsky, line 3: Pls update, there is no `blnd-tender`.

gromakovsky: If it was copy-pasted from somewhere, most likely their license requires you to preserve the copyright notice and the license.

backend/src/Edna/Config/Utils.hs
```diff
@@ -0,0 +1,22 @@
...
+import Edna.Setup (Edna, edConfig)
+
+configPathParser :: Opt.Parser (Maybe FilePath)
+configPathParser = Opt.optional $ Opt.strOption $
+  Opt.short 'c' <>
+  Opt.long "config" <>
+  Opt.metavar "FILEPATH" <>
+  Opt.help "Path to configuration file."
+
+fromConfig :: ((a -> Const a a) -> EdnaConfig -> Const a EdnaConfig) -> Edna a
```
gromakovsky: Is there no synonym for `(a -> Const a a) -> EdnaConfig -> Const a EdnaConfig`?

backend/src/Edna/Config/Utils.hs
```diff
@@ -0,0 +1,22 @@
...
+fromConfig getter = (^. edConfig . getter) <$> ask
```
gromakovsky: AFAIK it can be simplified to `view (edConfig . getter)`

backend/src/Edna/DB/Connection.hs
gromakovsky: Looks like a fail :smile:

backend/test/Test/Edna.hs
```diff
@@ -1,12 +1,18 @@
 module Test.Edna
   ( unit_stub_is_2
+  , spec_db
   ) where
 
 import Universum
 
+import Test.Hspec (Spec, it, shouldBe)
```
gromakovsky, line 8: Oh, I think we should use either `tasty` or `hspec` but not both. I previously tried both and I think both are acceptable, I liked `tasty` slightly more, but `hspec` is also ok. One advantage of `tasty` is that it can generate XML reports that are rendered nicely by GitLab, but here we are using GitHub, so it doesn't matter.

If you strictly prefer `hspec`, let's use it, then remove `tasty` completely. If you are ok with `tasty`, let's use it and not use `hspec`. Using both is redundant.

backend/stack.yaml
```diff
@@ -2,3 +2,16 @@ resolver: lts-17.2
...
+- beam-postgres-0.5.0.0
+- beam-migrate-0.5.0.0
+- dependent-map-0.4.0.0
+- dependent-sum-0.7.1.0
+- constraints-extras-0.3.0.2
+
+allow-newer: true
+
+
+
```
gromakovsky: Why so many empty lines? :smile:

gromakovsky (changes requested):
I did only superficial review, will follow-up soon.

backend/stack.yaml
pva701: gromak ti cho pyaniy? cho za 😆 pod kajdim messagem?

backend/src/Edna/Setup.hs
```diff
@@ -0,0 +1,28 @@
...
+import Edna.Config.Definition (EdnaConfig(..))
+import RIO (RIO, runRIO)
+
+data EdnaContext = EdnaContext
+  { _edConfig :: !EdnaConfig
+  }
+
+makeLenses ''EdnaContext
+
+type Edna = RIO EdnaContext
```
gromakovsky, line 25: Minor naming suggestion: maybe let's call it `EdnaM`? That sounds more consistent with other projects I have worked on. But maybe that's not the case for your projects, so feel free to ignore. It can have another suffix, e. g. `EdnaT`, just to be more specific than `Edna`.

backend/src/Edna/Setup.hs
```diff
@@ -1,28 +1,41 @@
...
   { _edConfig :: !EdnaConfig
+  , _edConnectionPool :: !ConnPool
   }
 
 makeLenses ''EdnaContext
 
 type Edna = RIO EdnaContext
 
+setEncoding :: IO ()
+setEncoding = do
```
gromakovsky: There is `with-utf8` library maintained by us, let's use it. Just wrap `main` into `withUtf8`.

backend/src/Edna/Util.hs
```diff
@@ -1,19 +1,33 @@
...
+  } deriving stock (Show, Eq, Ord)
+
+instance FromJSON ConnString where
+  parseJSON = withText "ConnString" $ pure . ConnString . encodeUtf8
+
+instance ToJSON ConnString where
+  toJSON = String . decodeUtf8 . unConnString
+
+instance Buildable ConnString where
+  build (ConnString s) = ""+|decodeUtf8 @Text s|+""
```
gromakovsky: `|+""` is likely needed, but `""+|` is not AFAIK. Also it may be simpler to write `build $ decodeUtf8 @Text s` (but that's up to you).

backend/src/Edna/DB/Integration.hs
```diff
@@ -0,0 +1,67 @@
...
+import Database.Beam.Schema (Beamable)
+import Database.PostgreSQL.Simple.Transaction (withTransactionSerializable)
+import RIO (withRunInIO)
+
+import Edna.DB.Connection (ConnPool(..))
+import Edna.Setup (Edna, edConnectionPool)
+
+withConnection :: (Connection -> Edna a) -> Edna a
+withConnection action = do
+  (ConnPool pool) <- (^. edConnectionPool) <$> ask
```
gromakovsky:
```suggestion
  ConnPool pool <- view edConnectionPool
```

I am not sure if it works, but I think it should.

backend/sql/initial_schema.sql
gromakovsky: > we didn't clean up it properly previous time (like there is a forgotten table in drop.sql)

@pva701 I don't understand, why do we want to clean up? Are you talking about developer workflow? I mean in the normal production workflow we don't need `drop.sql`, right? We run the production server, restart it for some reason (maybe the computer just died for example) and we actually want to `create table if not exists` in this case, don't we?

pva701: I meant developer workflow right.

>we actually want to create table if not exists in this case, don't we?

I don't know, you told himself that in production workflow we create db once, so why do we need this check `if not exists`?

pva701: I mean, this db creation from code on every run seems not to be done in normal flow. I mean, this is just an ugly hack for development which resided in disciplina/agora which considered de-facto standard in our web apps.

But what is the purpose of this code to be run on production? I mean, even if we take into consideration migrations, they rather will check some last applied sql migration and try to apply next, so no relying on existing table names, it's just nonsense.

This code should be run only in dev mode IMHO, and in devmode as schema might be changed often, it should be robust, but checking that there exists table with the same name doesn't seem robust to me (in most cases it will be previous version of table if the db was forgotten to be dropped).

backend/src/Edna/DB/Schema.hs
pva701: also, it doesn't seem to conform to migration approach (where there will be several sql files in a folder which will be applied sequentially, as I understand, so no compile time dependency).

But actually main disadvantage how painful it to recompile on every chih, so waiting for recompilation might be just annoying, it's a main reason. 

@gromakovsky che dumaesh? che ignorish.

backend/src/Edna/DB/Schema.hs
```diff
@@ -0,0 +1,23 @@
+module Edna.DB.Schema
```
gromakovsky, line 1: I am reading the [tutorial](https://haskell-beam.github.io/beam/tutorials/tutorial1/) and I see that it starts with defining types that hold data from tables, like `ProjectT` in our case for example. Are they supposed to be defined here? Just checking, I think it's ok to do that after [EDNA-27](https://issues.serokell.io/issue/EDNA-27).

backend/src/Edna/Web/Server.hs
gromakovsky: AFAIU this `ensureSchemaIsSetUp` doesn't actually ensure anything. We may see that `project` table exists and assume that everything is alright, but it may be a different `project` table, not the one we need (old format or even something completely unrelated that coincidentally was called `project`). AFAIU it only checks the name and nothing else:

> Do not throw an error if a relation with the same name already exists. A notice is issued in this case. Note that there is no guarantee that the existing relation is anything like the one that would have been created.

gromakovsky:
So AFAIU we have 2 related questions:
1. Should `IF NOT EXISTS` be used in `initial_schema.sql`?
2. Should we import `initial_schema.sql` in compile-time?

I would add one more which is kinda more general: should tables be created/dropped from the application or externally (e. g. have some small scripts)?

And AFAIU we have 3 workflows:
1. Development workflow: we may run the server often, sometimes we want to use an existing DB, sometimes we want to start from scratch. We (developers) can run the server always from the repo root (or `backend/`) and thus assume a certain location of .sql files.
2. Production workflow: the server is started once and restarted rarely, we usually want to use the existing DB, sometimes we want to migrate the data. If we expect some files to be present in some location, it should be handled by our packaging.
3. Staging workflow: something between development and production. The server is restarted on each push to some branch (I guess `master` for now), we probably want to run from scratch every time. If we expect some files to be present in some location, it should be handled by our packaging (like in production).

Currently we have no migrations and don't want to bother with them, so whenever we update the schema we will have to start from scratch.

Regarding `IF NOT EXISTS`, I think we should remove it because it really doesn't guarantee anything (as I wrote [here](https://github.com/serokell/edna/pull/8#discussion_r588180526)).

> 2. Should we import `initial_schema.sql` in compile-time?

Regarding this, I think the reason to embed it in compile-time is to make `edna-server` less dependent on the place you run it from. If this file is embedded, you can run it from anywhere.
In production and staging workflows I think it shouldn't really matter because AFAIU we can put all necessary files into OCI images and make them available in runtime.
In development workflow we will likely use `stack exec` and `cabal run` and I _guess_ they run the thing you ask them to run from the package root, even if you do `stack exec` from `src/Foo/Bar`, so it shouldn't matter here as well.

So it seems there is indeed no advantage in embedding `initial_schema.sql` in compile-time and there is one disadvantage mentioned by @pva701 (extra time spent on compilation). I don't think the disadvantage is too important because the schema shouldn't change too often, but it's still a disadvantage and there are no big advantages. So, unless @sashasashasasha151 has other reasons to embed the schema, let's read it in runtime instead.

> Should tables be created/dropped from the application or externally (e. g. have some small scripts)?

I didn't work on such apps before, so I don't know what's more convenient for dev workflow. But if we do it from the app, I think it should be a separate functionality, like a command or option. AFAIU there are two things we may want to do: clear the DB and create it from scratch. We can also have a combined option (recreate).
Regarding production and staging workflow I will ask Chris.

backend/scripts/test.sh
```diff
@@ -0,0 +1,36 @@
...
+# Running a server in background.
+# Set log level to warning to avoid polluting NOTICE messages in tests.
+conn_str="$(pg_tmp)&options=-c%20client_min_messages\%3DWARNING"
+
+# Executing tests.
+# This thing wraps every argument into double quotes because that's what stack wants :/
+args=( "${@/#/\"}" ); args=( "${args[@]/%/\"}" )
+TEST_PG_CONN_STRING="$conn_str" stack test --no-nix-pure --test-arguments "${args[*]}"
+
+# pg_tmp shutdowns and cleans up automatically after a period of inactivity
```
gromakovsky, line 36: Some extra comments about this file:
1. Is it supposed to be equivalent to what we are doing in `backend/default.nix`? AFAIU it doesn't look equivalent currently, e. g. we pass `-w 600` there and don't pass here. I think it'd be nice to unify as much as possible, though it's minor. `client_min_messages` is another difference.
2. It assumes `stack`, but I prefer `cabal`, so it'd be nice to make it work with `cabal` as well if possible. In `Makefile` we use `EDNA_USE_CABAL` variable to determine what to use. Since you are not using `cabal`, I guess I can try doing it myself. One option is to just call `make` and then expect `edna-test` to be in the `$PATH` and just call it directly. Currently `make` doesn't put it there, but probably it's doable. If we manage to put `edna-test` into `$PATH`, we can avoid weird code that wraps arguments passed to `--test-arguments`.
3. If all this stuff is necessary to make test-suite work, we should update `make test` as well. We can call this script from there or inline it there.

backend/scripts/test.sh
```diff
@@ -0,0 +1,36 @@
...
+stack test --no-run-tests
+
+# Running a server in background.
+# Set log level to warning to avoid polluting NOTICE messages in tests.
+conn_str="$(pg_tmp)&options=-c%20client_min_messages\%3DWARNING"
+
+# Executing tests.
+# This thing wraps every argument into double quotes because that's what stack wants :/
+args=( "${@/#/\"}" ); args=( "${args[@]/%/\"}" )
+TEST_PG_CONN_STRING="$conn_str" stack test --no-nix-pure --test-arguments "${args[*]}"
```
gromakovsky: 1. I am not sure if this wrapping is necessary, I never saw such code, can you please check whether it's really necessary and provide an example where it's necessary?
2. What is `--no-nix-pure`? I can't find it in `stack --help` and `stack build --help`. Maybe I am blind.

backend/scripts/test.sh
```diff
@@ -0,0 +1,36 @@
...
+
+set -e -o pipefail
+
+if ! which initdb > /dev/null; then
+    echo -e "Command 'initdb' not found.\nConsider adding '/usr/lib/postgresql/VERSION/bin/' to your PATH."
+    exit 2
+fi
+
+# A temporal server does not live long (60 sec by default), building tests first.
+stack test --no-run-tests
```
gromakovsky, line 25: I think it's better to use `make` here because it also passes `--fast` and may pass other flags we find essential for dev build.

gromakovsky (commented):
TODO: ensure that I can run tests on my machine.

backend/config.yaml
sashasashasasha151: So, I need to write smth like `host=/run/postgresql dbname=edna` or what?

backend/src/Edna/DB/Schema.hs
sashasashasasha151: >Are they supposed to be defined here?

Yes, in EDNA-38

backend/default.nix
sashasashasasha151: As I can see in docs yes

backend/scripts/test.sh
sashasashasasha151: I don't know the original place of copy-pasting it\
It was in disciplina/agora/blnd and so on (maybe before disciplina there also were some projects)\
So, from where I need to take it?

backend/src/Edna/Setup.hs
sashasashasasha151: So, it was called not by its haskell meaning but by common meaning\
But I can rename it if you want (with 🤔 about how many fixup commits I need to do it :peka:)

backend/scripts/test.sh
sashasashasasha151: 1. 600 is default value as I can see from comments. I don't know but maybe `client_min_messages` is different for CI for some reasons (maybe we do not need warnings). What do you think about it and what do you think we need to use?
2. I do not use `cabal` so you are welcome to do it. But if you want I can investigate how to do it by myself

backend/scripts/test.sh
sashasashasasha151: 2. https://docs.haskellstack.org/en/stable/nix_integration/#the-nix-shell

backend/scripts/test.sh
sashasashasasha151: Maybe in this case we can run tests from make and say to build before tests?

sashasashasasha151:
@gromakovsky 
> Regarding `IF NOT EXISTS`, I think we should remove it because it really doesn't guarantee anything

Maybe I don't understand smth but `IF NOT EXISTS` do not need to guarantee smth. It is just a hack to run script on every run and create tables id they are not present in DB.\
I read arguments why in some cases we can lose with `IF NOT EXISTS` but the probability is very small as for me.\
And for development it is very useful in some cases. For example it is normal case when you just remove some tables after fixing the code, rebuild it, simply run and continue testing. Without it will be needed to create these tables by hands.

>import initial_schema.sql in compile-time

It is how checking that schema is setup works, so if we do not want to run the script we can remove it

>Should tables be created/dropped from the application or externally

Ya uze zaputalsya v telegah, so will write my opinion. Let's add option to config which will run schema setup or not. For development we will run it as it is useful (I wrote above), but on production we will disable it and do all things manually. (also I believe that one day we will have migrations and all this telegas will be unrelated)

backend/scripts/test.sh
gromakovsky: Do you or anybody else use stack+nix integration? If someone does, okay, let's keep `--no-nix-pure` here, it's not a big deal.
I heard a number of times that stack+nix integration is quite bad, but if it works fine for you or somebody else, you are welcome to use it of course.

backend/config.yaml
gromakovsky: I think `host` and `dbname` are ok, the things that don't match are `user` and `password`. So I think you just need to change `user` to `ednauser` and `password` to `ednapass`.

gromakovsky:
> Let's add option to config which will run schema setup or not. 

Yes, I think everyone will be satisfied if we do it this way. I think we still don't need `IF NOT EXISTS` in this case, we can drop all tables before creating them. Because there are 2 cases: you want to use the existing DB which is already initialized, then you turn off DB initialization or you want to start with empty DB, then drop is harmless.

> It is how checking that schema is setup works, so if we do not want to run the script we can remove it

I think @pva701 suggests reading it in runtime to avoid recompilations when we change something in SQL files. Also it will remove `file-embed` dependency which is a minor advantage. I don't see any serious drawbacks if we read it in runtime. We can have `"init-db": "path/to/init.sql"` in the config and in the dev config it will simply always point to `init.sql` in this repo. In _production_ we can just bundle this `init.sql` and make sure it's available in runtime somewhere, it shouldn't be a problem according to Chris.

So to sum up these to points, I propose to add optional `"init-db": "path/to/init.sql"` to the config. If it's not present, we assume that DB is already initialized and start using it. If it's present, we read `init.sql` from that place and execute it. Before doing that we should also drop all tables. We have `drop.sql` for that, but apparently it can be done using just one command (see [EDNA-32](https://issues.serokell.io/issue/EDNA-32)), so hopefully we don't need this file.

gromakovsky:
>  Remove whitespaces 

@sashasashasasha151 btw, I am pretty sure the IDE you are using has an option to delete whitespaces, please enable it if it's not already enabled. Though maybe it's not even your whitespaces :thinking: Don't want to check :yao:

README.md
gromakovsky: Please move it to `backend` because it's about backend tests. I recently moved some information about backend to `backend/README.md`.

gromakovsky (changes requested):
Let's temporarily forget about all my comments regarding tests. I already have a task to update something in tests, so I'll take a fresh look at the tests within that task and will try to improve things that I don't like.

Summary of remaining changes, just to make it easier for me to review:
1. Add `init-db` to the config, make sure it's present in the default config because it's used by developers. Remove `IF NOT EXISTS` because it's a hack that won't be necessary anymore. In dev mode we'll just enable `init-db`.
2. Update user and password in the default connection string.
3. Make sure README updates go to `backend/`.

I don't mind if you do (3) in-place (force push). If (1) appears to interfere with many commits, it's ok to put it into a separate commit, even though it fixes work previously done in this PR. (2) sounds like a small fixup commit.

Also I recommend to rebase on master before making any new changes.

backend/scripts/test.sh
sashasashasasha151: >use stack+nix integration

I think no) Will remove

backend/src/Edna/Util.hs
```diff
@@ -55,6 +56,21 @@ instance FromJSON NetworkAddress where
...
 
+data DatabaseInitOption = Disable | Enable | EnableWithDrop
+  deriving stock (Generic, Show)
+
+instance FromJSON DatabaseInitOption where
+  parseJSON = withText "DatabaseInitOption" $ \o -> pure case o of
+    "disable" -> Disable
+    "enable" -> Enable
+    "enable-with-drop" -> EnableWithDrop
+    _ -> error $ "Invalid config value" <> show o
```
gromakovsky: `error $ "Invalid config value: " <> show o`

gromakovsky:
One more nitpick regarding commit messages :peka:
Please wrap lines to be at most 72 characters long.

image

README.md
```diff
@@ -25,6 +25,10 @@ Please refer to the respective directories for more details.
 To run a postgres container run `docker-compose up postgres` in `deployment` directory,
 then you can connect to it via `psql` or using any convenient database viewer.
 
+## Tests
+
+To run tests you need to have `pg_tmp` to run DB for tests. Then run `./scripts/test.sh` to execute tests.
```
gromakovsky: Move it to `backend/README.md` please.

gromakovsky (changes requested):
Two minor comments, can proceed with merging after that. I'll be around today, please ping me if you fix it today, I'll approve.

gromakovsky:
Also I told you in Slack how to fix CI issue.

backend/src/Edna/Config/Definition.hs
```diff
@@ -35,6 +37,8 @@ data ApiConfig = ApiConfig
 data DbConfig = DbConfig
   { _dbConnString :: ConnString
   , _dbMaxConnection :: Int
+  , _dbInitDb :: DatabaseInitOption
+  , _dbInitScript :: FilePath
```
gromakovsky: Oh, more more thing. I think it should actually be optional because if `init-db` is set to `"disable"` this file is ignored. So I think the UX will be better if the person who doesn't want to init db doesn't need to think whether they need to put something sane into `init-script`.
That said, I think our whole configuration approach is quite unstable and imperfect, we will refine it after MVP (or even during MVP development if it appears to cause too many troubles), so you can ignore this comment for now.
I think it makes sense to do it if it takes ≈5 minutes. If it takes more, it's not worth the effort.

gromakovsky (approved):

pva701:
LGTM, btw!

PR #9, opened 2021-03-01
Frontend checks on ci

sashasashasasha151:
## Related issue(s)

https://issues.serokell.io/issue/EDNA-19

- Tests
        silently reappearing again.

- Documentation


frontend/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "trailingComma": "es5",
+  "tabWidth": 2,
+  "semi": true,
+  "singleQuote": false,
+  "printWidth": 100,
+}
```
mkaito, line 7: Does this come from some style guide, or is this just your opinion?

flake.nix
```diff
@@ -13,21 +13,46 @@
...
         packages = {
           backend-lib = backend.library;
           backend-server = backend.server;
+          frontend = frontend.package;
         };
 
-        checks.backend-test = backend.test;
+        checks = {
+          backend-test = backend.test;
+          frontend-tscompile = frontendCheck "yarn run tscompile";
```
mkaito, line 52: Shouldn't this be part of the normal build process, or is this something else?

frontend/.prettierrc
sashasashasasha151: So we do not have Serokell style guide, but that one comes from tender app

mkaito (approved):

frontend/src/utils/request.ts
```diff
@@ -1,15 +1,15 @@
...
 // This file might be removed if we start using redux-query or recoil
 
-export type RequestState<T>
-  = { status: "idle" }
+export type RequestState<T> =
+  | { status: "idle" }
   | { status: "loading" }
-  | { status: "succeeded", result: T }
-  | { status: "failed", error?: string };
+  | { status: "succeeded"; result: T }
```
pva701, line 7: how I sick and tired of this bullshit. why is it even important? 

I get used to commas in types, you don't even write frontend but set up some rules how it should look like. why it matters? 

total crap, wasting of time.

nvm, it's just shit which I should swallow now and then I will revert it back when your attention will be focused on tight deadlines and not on this unessential miserable rules.

mkaito: That's actually just `prettier`. You can just set your editor up to prettify your code on save and forget about it forever.

frontend/src/variables.scss
```diff
@@ -0,0 +1,40 @@
+$px: 0.0625rem;
+
+//
+// Normalize SCSS Variables
```
pva701: is it necessary? we discussed it's not

frontend/src/utils/request.ts
pva701: I did it, but the main issue that my ide is highlighting  all eslint problems on fly and the code is just highlighted red, I am scared/confused, and I have to tolerate this while I end up with some compiled code, then it might be finally prettified. So it just makes code writing harder to me.

frontend/src/pages/upload/UploadPage.tsx
```diff
@@ -34,45 +32,45 @@ const UploadPage: FunctionComponent = (): ReactElement => {
...
-              <td>{v.signal}</td>
-            </tr>
-          ))}
-        </tbody>
-      </table>
-      )}
+        {uploadingStatus.status === "loading" && <div className="spinner">Uploading...</div>}
 
-        {uploadingStatus.status === "loading"
-      && <div className="spinner">Uploading...</div>}
```
pva701, line 72: why braces are needed? no, I can't tolerate it. let's remove it.

it's 100% convenient to write
```
{statement &&
<Component>}
```

TSX is really verbose and cumbersome itself, and syntax is not always so easy readable, why do we need this one more thing which makes it's even harder?

I suffer to match all these insane closing braces like )}}) and you add one more shitty rule?

I bet you suffer himself but you just following blindly to this cargo cult imposed by configs which are written by idiots from big companies like Airnbn and Uber

pva701:
https://github.com/serokell/edna/pull/9/commits/deff0d359f6e3dc0d61693aa0b47ffd7b5878428#r586582034

Can we remove this rule? Everything else is more or less fine

pva701 (approved):

PR #10, opened 2021-03-03
Initialise stub upload page

pva701:
## Related issue(s)

https://issues.serokell.io/issue/EDNA-24

- Tests
        silently reappearing again.

- Documentation


frontend/.eslintrc.js
```diff
@@ -48,6 +48,7 @@ module.exports = {
     // Use TS no-unused-vars rule
     "no-unused-vars": "off",
     "@typescript-eslint/no-unused-vars": ["error"],
+    "@typescript-eslint/explicit-module-boundary-types": 0,
```
sashasashasasha151: Why?

frontend/src/pages/upload/UploadPage.tsx
```diff
@@ -1,58 +1,152 @@
...
-              uploadInputRef.current &&
-              uploadInputRef.current.files &&
-              uploadInputRef.current.files.length > 0
-            ) {
-              setUploadingStatus(loading());
-              const file = uploadInputRef.current.files[0];
-              const measurements = await Api.uploadExperiment(file);
-              setUploadingStatus(completed(measurements));
+    <>
+      <Header />
```
sashasashasasha151: I think it is time to create `Router` and some `EdnaPage` which will contain header and footer (in the future) and take your content as a child. Just to formulate common layout of all pages.

frontend/src/components/Header/Header.scss
```diff
@@ -0,0 +1,51 @@
...
+@import "src/styles/variables";
+
+.ednaHeader {
+
+  $menuitem-border: 0.3125rem;
+  display: flex;
+  align-items: center;
+  height: 84 * $px;
+  padding-right: $global-side-padding;
+  padding-left: $global-side-padding;
```
sashasashasasha151: Is it discussed solution? So on wider screens header will be wider then on tight screens?

sashasashasasha151:
```suggestion
  padding: 0 $global-side-padding;
```

frontend/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <div className="ednaHeader">
+      <span className={ednaHeader("brand")}>Edna</span>
+
+      <div className={ednaHeader("menu")}>
+        {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+        <a className={ednaHeader("menuitem")}>Dashboard</a>
+
+        {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+        <a className="ednaHeader__menuitem ednaHeader__menuitem_active">Upload</a>
```
sashasashasasha151: You can add active link to props. (but maybe it was just an example)\
Also why not to use `ednaHeader()` here?

frontend/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,26 @@
...
+const Header: FunctionComponent = (): ReactElement => {
+  const ednaHeader = cn("ednaHeader");
+
+  return (
+    <div className="ednaHeader">
+      <span className={ednaHeader("brand")}>Edna</span>
+
+      <div className={ednaHeader("menu")}>
+        {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+        <a className={ednaHeader("menuitem")}>Dashboard</a>
```
sashasashasasha151: Just create list of menuItems instead of writing them 3 times

frontend/src/components/Header/Header.scss
```diff
@@ -0,0 +1,51 @@
...
+    height: 100%;
+    margin-left: auto;
+  }
+
+  &__menuitem {
+    $wrapper-padding: 10 * $px;
+    $wrapper-margin: 30 * $px;
+    display: flex;
+    flex-direction: column;
+    justify-content: space-around;
```
sashasashasasha151: :pogchamp: but do you need space to the right of the last item? Now spaces on the left and on the right are different and on design they are same

frontend/src/components/Header/Header.scss
```diff
@@ -0,0 +1,51 @@
...
+    height: 100%;
+
+    padding-right: $wrapper-padding;
+    padding-left: $wrapper-padding;
+    margin-right: $wrapper-margin;
+    margin-left: $wrapper-margin;
```
sashasashasasha151: But you already have `justify-content: space-around;` \
So, I think it will be better to remove `justify-content` and add `& + & { margin-left: <size> }`

frontend/src/pages/upload/UploadPage.tsx
```diff
@@ -1,58 +1,152 @@
...
+      <div className="container">
```
sashasashasasha151: Move to separate component together with `Header`

frontend/src/utils/utils.ts
```diff
@@ -0,0 +1,10 @@
+// For testing and mocking purposes
+export function delay(ms: number) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+export type Maybe<T> = T | undefined;
+
+export function isDefined<T>(anyVal: T | undefined | null): anyVal is T {
+  return anyVal !== undefined && anyVal !== null;
+}
```
sashasashasasha151, line 10: Maybe I don't understand smth but what for do you need it? `file: Maybe<File>;` can be `file?: File;` and `<FormField<Maybe<Project>>` can be `<FormField<Project | undefined>` \
And `if (!isDefined(values.file)) {`  `if (!values.file) {` \

Can you please explain where I am wrong?

frontend/src/pages/upload/UploadPage.tsx
```diff
@@ -1,58 +1,152 @@
...
+  description: string;
+}
 
 const UploadPage: FunctionComponent = (): ReactElement => {
-  const uploadInputRef = useRef<HTMLInputElement | null>(null);
   const [uploadingStatus, setUploadingStatus] = useState<RequestState<MeasurementDto[]>>(idle());
+  const createProject = createProjectUpdater();
+  const createMethodology = createMethodologyUpdater();
+  const projectsLoadable = useRecoilValueLoadable(projectsAtom);
+  const methodologiesLoadable = useRecoilValueLoadable(methodologiesAtom);
```
sashasashasasha151, line 31: Am I right that it is like `useSelector` in `redux`?

frontend/src/store/atoms.ts
```diff
@@ -0,0 +1,20 @@
...
+import { atom, RecoilState, selector } from "recoil";
+import { Methodology, Project } from "./types";
+import Api from "../api/api";
+
+export const projectsAtom: RecoilState<Project[]> = atom({
+  key: "Projects",
+  default: selector({
+    key: "Projects/Default",
+    get: () => Api.fetchProjects(),
+  }),
```
sashasashasasha151, line 11: So, on app start it will fetch projects and methodologies, right? Is it possible to configure what thins we need to load for specific page? For example for one page we need only projects and for other only methodologies, so it not needed to load both of them

frontend/src/pages/upload/UploadPage.tsx
sashasashasasha151: I also see that there are selectors in recoil, what is the difference

frontend/src/pages/upload/UploadPage.scss
```diff
@@ -1,10 +1,11 @@
...
 
 .uploadPageContainer {
   display: flex;
   flex-direction: column;
   align-items: center;
-  max-width: 43.75rem;
+  max-width: 700 * $px;
   margin-right: auto;
   margin-left: auto;
 }
```
sashasashasasha151: You do not use it

frontend/src/pages/upload/UploadPage.tsx
```diff
@@ -1,58 +1,152 @@
...
+              tabIndex={4}
+            />
+
+            <button className="ednaButton uploadingForm__submitBtn" type="submit" tabIndex={5}>
+              Save
+            </button>
+          </Form>
+        </Formik>
+
+        <div className="uploadPageResult">
```
sashasashasasha151: You use this class here and in `MeasurementsCharts`

frontend/src/components/UploadArea/UploadArea.tsx
```diff
@@ -0,0 +1,44 @@
...
+type UploadingAreaProps = FormikCompatible<File>;
+
+const UploadArea: FunctionComponent<UploadingAreaProps> = ({ value, onChange }): ReactElement => {
+  const uploadInputRef = useRef<HTMLInputElement | null>(null);
+
+  return (
+    <>
+      <input
+        ref={uploadInputRef}
+        style={{ display: "none" }}
```
sashasashasasha151: Add classname

sashasashasasha151 (changes requested):

frontend/src/pages/upload/UploadPage.tsx
pva701: it's done in another branch, about library mockup

frontend/.eslintrc.js
pva701: perhaps, due to it's too silly to specify React.ReactElement for each component type, and also for some one-liner functions. perhaps, can be reverted

frontend/src/components/Header/Header.scss
pva701: perhaps, it makes sense to wrap a header in some container instead of having padding

frontend/src/components/Header/Header.tsx
pva701: this part is affected/fixed in Library mockup pr (as there is a Router)

frontend/src/components/Header/Header.tsx
pva701: it's just an example, also let's move the discussion in another PR where it's affected

frontend/src/components/Header/Header.scss
pva701: you confuse things, justify-content is to vertical axes but margin to horizontal. I will do it center instead of space-around perhaps

frontend/src/pages/upload/UploadPage.tsx
pva701: together with you ass :putya:

frontend/src/utils/utils.ts
pva701: lol, it's just convenient alias, it's more convenient for me to write `isDefined` and don't be worried that somewhere I treated empty string, empty list or whatever as `false` where I didn't want to

frontend/src/pages/upload/UploadPage.tsx
pva701: no, it's not like useSelector. 
it's like `useState` in React (but without the setter) and also wrapped in `Loadable` wrapper (when it's loading corresponding `hasValue` will be returned)

frontend/src/store/atoms.ts
pva701: lol, it will do exactly opposite: when you access to it at the first time it will start fetching the data, not on start. it's possible to make prefetch to start loading on start.

frontend/src/styles/main.scss
```diff
@@ -31,13 +31,12 @@ $global-side-padding: 138 * $px;
   box-shadow: 0 0 0 $px $input-border-outlined;
 }
 
-.ednaButton {
+.primaryButton {
```
pva701, line 35: this buttons stuff is moved to a separate component in further branch btw :putya:

frontend/src/utils/utils.ts
sashasashasasha151: It is just :ghc: for people who read it

frontend/src/pages/upload/UploadPage.tsx
sashasashasasha151: So, it will not change when you will set different value to it? Or how? Also what about second question?

frontend/src/components/Header/Header.scss
pva701: `& + & { margin-left: <size> }` this is also done

sashasashasasha151 (approved):

frontend/src/utils/utils.ts
pva701: it's just like `Maybe`: you rule out only `undefined` and `null` values, but not empty ones. also it's called type guards, dunno if it's the thing confusing you.

frontend/src/pages/upload/UploadPage.tsx
pva701: it will change if you set

selectors here and in redux are basically the same: they represent derived state

PR #11, opened 2021-03-04
[EDNA-14] Add OCI image packages

mkaito:
## Description

OCI images for backend and frontend containers. Built but not (yet) pushed in CI.

Build locally with `nix build .#docker-backend`.

Import [as documented](https://docs.docker.com/engine/reference/commandline/import/#import-from-a-local-file), then run the container. It expects environment variables `$API_PORT` and `$SERVE_DOCS` set.

Config layering is tentative until I figure out how exactly the backend needs to be configured.

Related PR: https://github.com/serokell/gemini-infra/pull/4

## Related issue(s)

https://issues.serokell.io/issue/EDNA-14

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -14,12 +14,20 @@ steps:
   - label: backend test
     commands:
       - nix-build -A checks.x86_64-linux.backend-test
+  - label: backend image
+    commands:
+      - nix-build -A packages.x86_64-linux.docker-backend
+      - buildkite-agent artifact upload "$(readlink -f result)"
   - label: backend hlint
     commands:
       - nix-shell --run 'hlint backend -j'
   - label: frontend build
     commands:
       - nix-build -A packages.x86_64-linux.frontend
+  - label: frontend image
+    commands:
+      - nix-build -A packages.x86_64-linux.docker-frontend
+      - buildkite-agent artifact upload "$(readlink -f result)"
   - label: frontend tscompile
     commands:
       - nix-build -A checks.x86_64-linux.frontend-tscompile
@@ -29,3 +37,8 @@ steps:
   - label: frontend stylelint
     commands:
       - nix-build -A checks.x86_64-linux.frontend-stylelint
+  - wait
+  - label: Deploy staging
+    commands:
+      - nix-shell --run 'deploy .#castor'
+    branches: "master"
--- a/docker.nix
+++ b/docker.nix
@@ -1,0 +1,84 @@
+{ dockerTools
+, backend
+, frontend
+, linkFarm
+, runCommand
+, writeTextDir }:
+
+let
+  inherit (dockerTools) buildLayeredImage buildImage pullImage;
+in
+
+{
+  backend-image = let
+    # linkFarm seems to put a dangling symlink in the image when used here, so
+    # we use runCommand and copy the file instead.
+    init-sql = runCommand "edna-init-sql" {} ''
+      mkdir -p $out
+      cp ${./backend/sql/init.sql} $out/init.sql
+    '';
+  in buildLayeredImage {
+    name = "ghcr.io/serokell/edna-backend";
+    tag = "latest";
+
+    contents = [
+      backend
+      init-sql
+    ];
+
+    config = {
+      Entrypoint = "/bin/edna-server";
+    };
+  };
+
+  frontend-image = let
+    nginx = pullImage {
+      imageName = "nginx";
+      imageDigest = "sha256:e20c21e530f914fb6a95a755924b1cbf71f039372e94ac5ddcf8c3b386a44615";
+      sha256 = "1xfl1yx5mklxzcbywmfr8y48kkkp7if364r24ghjz33jkb3wm8vy";
+      finalImageName = "nginx";
+      finalImageTag = "alpine";
+    };
+    html = linkFarm "edna-frontend-bundle" [{
+      name = "usr/share/nginx/html";
+      path = "${frontend}";
+    }];
+    nginx-conf-template = writeTextDir "etc/nginx/templates/edna.conf.template" ''
+      server {
+        listen 80 default_server;
+        root /usr/share/nginx/html;
+        location / {
+          try_files $uri /index.html =404;
+        }
+
+        location ''${API_PATH} {
+          proxy_pass http://''${API_HOST}:''${API_PORT}/;
+        }
+      }
+    '';
+  in buildImage {
+    name = "ghcr.io/serokell/edna-frontend";
+    tag = "latest";
+    fromImage = nginx;
+    contents = [
+      html
+      nginx-conf-template
+    ];
+
+    config = {
+      # Somehow, all this gets lost when inheriting from the parent image.
+      Entrypoint = [ "/docker-entrypoint.sh" ];
+      Cmd = [ "nginx" "-g" "daemon off;" ];
+      ExposedPorts = {
+        "80/tcp" = {};
+      };
+
+      # Used in the conf template above
+      Env = [
+        "API_PATH=/api/"
+        "API_HOST=backend"
+        "API_PORT=9000"
+      ];
+    };
+  };
+}
--- a/flake.nix
+++ b/flake.nix
@@ -3,6 +3,10 @@
     nixpkgs.url = "github:serokell/nixpkgs";
     haskell-nix.url = "github:input-output-hk/haskell.nix";
     flake-utils.url = "github:numtide/flake-utils";
+    serokell-nix.url = "github:serokell/serokell.nix";
+
+    deploy-rs.url = "github:serokell/deploy-rs";
+    deploy-rs.inputs.nixpkgs.follows = "nixpkgs";
 
     hackage = {
       url = "github:input-output-hk/hackage.nix";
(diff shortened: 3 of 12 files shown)


flake.nix
```diff
@@ -20,38 +21,46 @@
...
-        pkgs = nixpkgs.legacyPackages.${system}.extend
-          (foldl' composeExtensions (_: _: { }) [
-            nix-npm-buildpackage.overlay
-            haskell-nix.overlay
-          ]);
+        pkgs = pkgsWith nixpkgs.legacyPackages.${system} [
+          nix-npm-buildpackage.overlay
+          haskell-nix.overlay
+          serokell-nix.overlay
+        ];
```
gromakovsky, line 45: Oh nice, looks more understandable this way.

docker.nix
```diff
@@ -0,0 +1,99 @@
...
+  });
+
+  backend-entry-point = writeScriptBin "entrypoint.sh" ''
+    #!/bin/bash
+    set -euxo pipefail
+
+    /bin/envsubst '$API_PORT $SERVE_DOCS' \
+      < ${backend-config-template} >| /config.yaml
+
+    exec /bin/edna-server -c -c /config.yaml "$@"
```
gromakovsky: Looks like redundant duplicated `-c`.

docker.nix
```diff
@@ -0,0 +1,99 @@
...
+        if {path} not_starts_with /api/
+        if {path} not_starts_with /static/
+        to /
+      }
+    }
+  '';
+in
+{
+  backend-image = buildLayeredImage {
+    name = "ghcr.io/serokell/edna-backend";
```
gromakovsky, line 21: Is it the place where the image will be uploaded?

docker.nix
mkaito: I have no idea how it is run or configured yet :stuck_out_tongue: I need to have a look at that.

docker.nix
mkaito: Well, yes. But also, no. The image name must match the registry URI where it will be available. This name does not necessarily imply that it is being pushed there now. It's just a name.

balsoft: I don't think this name _must_ match the registry URI.

docker.nix
```diff
@@ -0,0 +1,99 @@
...
+  # Some system files necessary for nginx to function normally
+  shadow-files = runCommand "shadow-files" {} ''
+    mkdir -p $out/etc
+    echo "nobody:x:65534:65534:nobody:/:/sbin/nologin" > $out/etc/passwd
+    echo "nobody:!x:::::::" > $out/etc/shadow
+    echo "nobody:x:65534:" > $out/etc/group
+    echo "nobody:x::" > $out/etc/gshadow
+  '';
+
+  backend-config-template = writeText "backend-config-env.yaml.template" (toJSON {
```
balsoft: BTW you can import the `config.yaml` that's in the repo here and merge it with those overrides. See https://gitlab.com/tezosagora/agora/-/blob/master/nix/modules/services/backend.nix#L61 for example

docker.nix
```diff
@@ -0,0 +1,99 @@
...
+    buildLayeredImage
+    pullImage
+    shadowSetup;
+
+  inherit (builtins) toJSON;
+
+  httpPort = 80;
+  httpsPort = 443;
+
+  # Some system files necessary for nginx to function normally
```
balsoft: nginx?

docker.nix
mkaito: Oh, nice!

mkaito:
The following environment variables are available at runtime in the backend container:

          POSTGRES_HOST 
          POSTGRES_DB
          POSTGRES_USER
          POSTGRES_PASSWORD

I can interpolate them into a config file, or you can consume them from the environment once you get around to postgres usage.

The staging server's canonical name is `castor.gemini.serokell.team`. It also has a CNAME `staging.edna.serokell.team` so you don't have to remember it.

I'm still fiddling with a few runtime things, but feel free to review.

.buildkite/pipeline.yml
```diff
@@ -11,12 +11,18 @@ steps:
...
+      - nix-build -A packages.x86_64-linux.docker-backend
   - label: backend hlint
     commands:
       - nix-shell --run 'hlint backend -j'
   - label: frontend build
     commands:
       - nix-build -A packages.x86_64-linux.frontend
+  - label: frontend image
+    commands:
+      - nix-build -A packages.x86_64-linux.docker-frontend
```
gromakovsky, line 29: I propose to add built images into CI artifacts to be able to quickly check what was built on CI.

mkaito: Sure, why not.

gromakovsky:
image
I am a Docker newbie :grimacing: So I imported 2 `tar.gz` files (not sure if correctly btw) and got 2 hashes. What should I do next to run them?

balsoft:
You shouldn't `docker import` those images, but rather `docker load -i` them. Then you'll be able to `docker run ghcr.io/serokell/edna-frontend` and `docker run ghcr.io/serokell/edna-backend`

mkaito:
Yeah, sorry. I'm not sure if the command semantics changed or my memory got corrrupted since the last time I worked with this kind of thing. @balsoft is correct.

frontend/default.nix
```diff
@@ -1,15 +1,21 @@
...
-      sha256 = "0dl91l414na44h090cgghd06q0j2whlj9h98im2qb9823glq7xff";
-    };
-    buildInputs = [ python2 ];
-
-    src = ./.;
-    yarnBuild = "yarn";
+
+buildYarnPackage {
+  SASS_BINARY_PATH = fetchurl {
+    url = "https://github.com/sass/node-sass/releases/download/v4.12.0/linux-x64-64_binding.node";
```
zhenyavinogradov: In the yarn lockfile it's `v4.14.1` instead of `v4.12.0`, I think it should be the same version here

gromakovsky:
@mkaito we have merged 2 PRs related to DB integration, so now 2 your PRs (this one and in another repo) need to be updated a bit. Specifically:
* `edna-server` now connects to a PostgreSQL DB. This is how it's configured:
```
db:
  conn-string: "host=localhost dbname=ednadb user=ednauser password=ednapass"
  max-connection: 200
```
`max-connection: 200` is probably ok.
* Also there is an optional configuration setting to init the DB:
```
  # Optional value used if you want to init DB on start
  initiation:
    # Two options of DB initialization from the app:
    # * "enable" will run init script on the app start,
    # * "enable-with-drop" will run drop script before init
    mode: "enable"
    init-script: "./sql/init.sql"
```

For staging, I think we can create DB from scratch on each push to master, so that we don't need to care about schema changes and migrations for now. So I think in the staging config `mode` should be `"enable-with-drop"` and you should make sure that `init-script` points to https://github.com/serokell/edna/blob/1a98cf6a37da384f7c23180c9db132e82e030803/backend/sql/init.sql

Is it ok, can you do it?

flake.nix
```diff
@@ -20,59 +24,109 @@
...
-        frontendCheck = checkPhase:
-          frontend.package.overrideAttrs (_: {
-            doBuild = false;
-            doCheck = true;
-            inherit checkPhase;
-          });
+    deploy.nodes.castor =
+      let
+        # pkgs.symlinkJoin et al can only handle folders, not files; so we roll
+        # our own, but skip the `dirname` call.
```
zhenyavinogradov: There is `linkFarm`, which can handle files

zhenyavinogradov (approved):

docker.nix
```diff
@@ -0,0 +1,75 @@
...
+      backend
+      entry-point
+
+      iana_etc
+      busybox
+    ];
+
+    config = {
+      Entrypoint = "/bin/entrypoint.sh";
+      Cmd = [ "-c" "/config.yaml" ];
```
gromakovsky: Where does it come from? AFAIU this `/config.yaml` is a path inside the OCI image that we are creating here. In which place do we create this file in the image?

flake.nix
mkaito: I tried `linkFarm` and it complained about it being a file.

docker.nix
mkaito: You're expected to provide it externally via mount or volume.

To provide more context, `Cmd` are given as arguments to `Entrypoint`. When you `docker run foo:latest banana`, `banana` overrides the `Cmd` portion. So it's just a default.

frontend/default.nix
mkaito: Updated

docker.nix
gromakovsky: So AFAIU the relevant part of `edna.nix` (in another repo) is this:
```
        volumes = [
          "${configFile}:/config.yaml:ro"
        ];

        cmd = [
          "-c" "/config.yaml"
        ];
```

flake.nix
mkaito: Nevermind, seems I'm just dumb.

docker.nix
mkaito: Yes. I've decided to remove the default Cmd from this repo, since it depends on where the user mounts the config file anyway.

gromakovsky (approved):
Looks plausible.

docker.nix
```diff
@@ -0,0 +1,84 @@
...
+  inherit (dockerTools) buildLayeredImage buildImage pullImage;
+in
+
+{
+  backend-image = let
+    # linkFarm seems to put a dangling symlink in the image when used here, so
+    # we use runCommand and copy the file instead.
+    init-sql = runCommand "edna-init-sql" {} ''
+      mkdir -p $out
+      cp ${./backend/sql/init.sql} $out/init.sql
```
gromakovsky, line 18: Minor suggestion: maybe let's put it into `$out/sql/init.sql` so that `dev-config.yaml` works? In `dev-config.yaml` we have `init-script: "./sql/init.sql"`.

PR #12, opened 2021-03-05
Parser based on sample files

sashasashasasha151:
Problem: we have sample files and need to parse them to show data and plots

Solution: create proper parser; add columns to chart on frontend to show target and outliers; use log for x axis on plots

## Related issue(s)

https://issues.serokell.io/issue/EDNA-26

- Tests
        silently reappearing again.

- Documentation


backend/src/Edna/ExperimentReader/Types.hs
```diff
@@ -0,0 +1,41 @@
...
+  ) where
+
+import Universum
+
+import Codec.Xlsx (CellValue(..))
+import Fmt (Buildable(..))
+
+data ParameterType = Target | Compound
+
+data Parameter = Parameter
```
pva701, line 25: Comment would be appreciated

backend/src/Edna/ExperimentReader/Parser.hs
```diff
@@ -0,0 +1,114 @@
...
+
+    _ -> Left $ InvalidCell position
+
+-- | Check that cell representation equals to some value
+cellSatisfy :: Worksheet -> Text -> (Int, Int) -> Bool
+cellSatisfy workSheet value position = case workSheet ^? ixCell position . cellValue of
+  Just (Just x) -> pretty x == value
+  _ -> value == ""
+
+-- | Common function for splitting
```
pva701: What splitting is? Function argument comments?

backend/src/Edna/ExperimentReader/Error.hs
```diff
@@ -0,0 +1,27 @@
...
+  | WorksheetNotFound
+  | TabletStartNotFound
+  | FileParsingError ParseError
+  | NoConcentrationTablet
+  deriving stock (Show, Generic, Eq)
+
+instance Buildable ExperimentParsingError where
+  build UnexpectedCellType = "Expected one type of cell but got another one"
+  build EmptyCell = "Cell is empty"
+  build (InvalidCell e) = "Undefined content of the cell" +| (show e :: String) |+ ""
```
gromakovsky: Btw, there is `||+`.

backend/src/Edna/ExperimentReader/Error.hs
```diff
@@ -0,0 +1,27 @@
...
+
+import Codec.Xlsx (ParseError)
+import Fmt (Buildable(..), (+|), (|+))
+
+data ExperimentParsingError
+  = UnexpectedCellType
+  | InvalidCell (Int, Int)
+  | EmptyCell
+  | WorksheetNotFound
+  | TabletStartNotFound
```
gromakovsky: By tablet do you mean plate? If I understand correctly what you mean, I propose to rename it to "plate" because that's what we use in our glossary at least.

backend/src/Edna/ExperimentReader/Parser.hs
```diff
@@ -0,0 +1,114 @@
...
+      restParams <- constructParams lastId (id2 : ids)
+      pure $ Parameter name (id1, id2 - 1) : restParams
+
+parseExperimentXls :: L.ByteString -> ParserType [ExperimentalMeasurement]
+parseExperimentXls content = do
+  xlsx <- left FileParsingError $ toXlsxEither content
+  workSheet <- maybeToRight WorksheetNotFound $ xlsx ^? xlSheets . ix 0 . _2
+
+  -- Check that tablet exists and start from the top left corner of the table
+  bool (Left TabletStartNotFound) (Right ()) $ cellSatisfy workSheet "<>" (1, 1)
```
gromakovsky: I think it's more understandable to write
```
  unless (cellSatisfy workSheet "<>" (1, 1)) $ Left TabletStartNotFound
```

backend/src/Edna/ExperimentReader/Parser.hs
```diff
@@ -0,0 +1,114 @@
...
+      name <- specificCellAt workSheet (direction (normalAxisPoint, idx)) CText
+      pure $ [Parameter name (idx, lastId)]
+    constructParams lastId (id1 : id2 : ids) = do
+      name <- specificCellAt workSheet (direction (normalAxisPoint, id1)) CText
+      restParams <- constructParams lastId (id2 : ids)
+      pure $ Parameter name (id1, id2 - 1) : restParams
+
+parseExperimentXls :: L.ByteString -> ParserType [ExperimentalMeasurement]
+parseExperimentXls content = do
+  xlsx <- left FileParsingError $ toXlsxEither content
```
gromakovsky: AFAIU `left` is available in `base` (`Control.Arrow`), there is no need to depend on `optparse-applicative` in the library just for `left`. Also there is `first` in `Universum`, it can be used instead of `left`.

gromakovsky (changes requested):

backend/src/Edna/ExperimentReader/Error.hs
```diff
@@ -0,0 +1,27 @@
...
+
+import Universum
+
+import Codec.Xlsx (ParseError)
+import Fmt (Buildable(..), (+|), (|+))
+
+data ExperimentParsingError
+  = UnexpectedCellType
+  | InvalidCell (Int, Int)
+  | EmptyCell
```
gromakovsky: Why do we attach cell location in `InvalidCell`, but do not attach it in `EmptyCell` and `UnexpectedCellType`?

backend/src/Edna/ExperimentReader/Parser.hs
```diff
@@ -0,0 +1,114 @@
...
+    (_, Nothing) -> Left EmptyCell
+    (_, Just (CellText "")) -> Left EmptyCell
+
+    (CText, Just (CellText x)) -> Right x
+    (CDouble, Just (CellDouble x)) -> Right x
+
+    (CSignal, Just (CellDouble x)) -> Right $ Signal x False
+    (CSignal, Just (CellText
+      -- If signal value is in brackets then it is outlier
+      (readParen True reads . unpack . replace "," "." -> ((d, "") : _))
```
gromakovsky: IMO using `toString` instead of `unpack` is better because in this case the type of conversion is clearer (there are many `unpack` functions), but it's really minor and I don't insist on doing that.

backend/src/Edna/Web/Handlers.hs
```diff
@@ -1,22 +1,21 @@
+{-# LANGUAGE StarIsType #-}
```
gromakovsky: This addition is somewhat weird because you didn't add any new code to this module, you only removed some :suspect:

backend/src/Edna/ExperimentReader/Error.hs
```diff
@@ -4,24 +4,38 @@ module Edna.ExperimentReader.Error
...
+
+deriving stock instance Show ExperimentParsingError
+instance Eq ExperimentParsingError where
+  UnexpectedCellType a b c == UnexpectedCellType d e f = a == d && c == f && Just b == (cast e)
+  InvalidCell a == InvalidCell b = a == b
+  EmptyCell == EmptyCell = True
+  WorksheetNotFound == WorksheetNotFound = True
+  FileParsingError a == FileParsingError b = a == b
+  NoConcentrationTablet == NoConcentrationTablet = True
+  _ == _ = False
```
gromakovsky: There is a chance that someone will add a new error constructor and forget to write `NewConstr == NewConstr = True` here. To avoid that, we can write `2 * n` cases instead of `n + 1`:
```
FileParsingError a == FileParsingError b = a == b
FileParsingError _ == _ = False
NoConcentrationTablet == NoConcentrationTablet = True
NoConcentrationTablet == _ = False
…
```

I propose to do it, it's not still linear number of cases, not quadratic, so shouldn't be a big problem. The idea is to avoid `_ == _` case.

backend/src/Edna/ExperimentReader/Types.hs
```diff
@@ -4,12 +4,18 @@ module Edna.ExperimentReader.Types
...
 
 import Universum
 
+import qualified GHC.Show as S
+
 import Codec.Xlsx (CellValue(..))
-import Fmt (Buildable(..))
+import Fmt (Buildable(..), (+|), (|+), pretty)
+
+data PointYX = PointYX (Int, Int)
```
gromakovsky: I would define it as
```
data PointYX = PointYX { pointY :: Int, pointX :: Int }
```

and then probably you can even drop `YX` from the name, but I don't insist on doing that, it's minor.

backend/test/Test/Parser.hs
```diff
@@ -0,0 +1,89 @@
...
+  let compounds = S.elems . S.fromList . map emCompoundId <$> parsingResult
+  let targets = S.elems . S.fromList . map emTargetId <$> parsingResult
+  let outliers = length . filter emOutlier <$> parsingResult
+  let numberOfMeasurements = length <$> parsingResult
+  targets `shouldBe` Right eTargets
+  compounds `shouldBe` Right eCompounds
+  outliers `shouldBe` Right eOutliers
+  numberOfMeasurements `shouldBe` Right eNumberOfMeasurements
+
+spec_parser :: Spec
```
gromakovsky, line 41: See my comment about `hspec` here: https://github.com/serokell/edna/pull/8#discussion_r587610487

I planned to use `tasty` and I think it doesn't make any sense to use both `hspec` and `tasty`. If you prefer `hspec`, that's fine, let's use `hspec`, but then please remove `tasty` from the codebase. Or let's use `tasty` and build `TestTree`s.

backend/test/Test/Parser.hs
```diff
@@ -0,0 +1,89 @@
...
+
+tablesPath :: FilePath
+tablesPath = "./resources/testSheets"
+
+getWorkSheet :: FilePath -> Text -> IO Worksheet
+getWorkSheet filePath sheetName = do
+  file <- L.readFile filePath
+  maybe (throwM WorksheetNotFound) pure $ (toXlsx file) ^? ixSheet sheetName
+
+tableBaseTest :: FilePath -> Text -> [Text] -> [Text] -> Int -> Int -> IO ()
```
gromakovsky: I think it's semantically more correct to write `Expectation` here instead of `IO ()` (they are synonyms).

backend/test/Test/Parser.hs
```diff
@@ -0,0 +1,89 @@
...
+tableBaseTest path sheetName eCompounds eTargets eOutliers eNumberOfMeasurements = do
+  workSheet <- getWorkSheet (tablesPath <> path) sheetName
```
gromakovsky: 1. You can use `</>` from `System.FilePath` here and avoid placing `/` in front of each `path` that you pass here.
2. Ideally you should avoid all `/` in this file because one day someone may want to run these tests on Windows for whatever reason. It's a bad style to hardcode `/` as path delimiter. 

I can accept usage of `/` (and discourage it) because it's unlikely that anybody will run it on Windows (but in theory Haskell works fine on Windows and for example someone can develop there), but let's at least insert `/` here rather than outside of this function.

backend/src/Edna/ExperimentReader/Error.hs
gromakovsky: Note that `EmptyCell` still doesn't include `PointYX` for unclear reason.

backend/src/Edna/ExperimentReader/Error.hs
```diff
@@ -0,0 +1,39 @@
...
+  PlateStartNotFound == PlateStartNotFound = True
+  _ == _ = False
+
+instance Buildable ExperimentParsingError where
+  build (UnexpectedCellType point expected actual) =
+    "Expected \"" +| expected |+ "\" cell type, but found " +| actual |+ " at " +| point |+ ""
+  build EmptyCell = "Cell is empty"
+  build WorksheetNotFound = "Worksheet not found"
+  build (FileParsingError e) = show e
+  build PlateStartNotFound = "\"<>\" plate start poin not found"
```
gromakovsky:
```suggestion
  build PlateStartNotFound = "\"<>\" plate start point not found"
```

backend/src/Edna/ExperimentReader/Parser.hs
```diff
@@ -0,0 +1,123 @@
...
+type ParserType a = Either ExperimentParsingError a
+
+-- | Based on specific CellType return its value or get and exception that cell is empty or its
+-- expected type does not satisfy actual type
+specificCellAt
+  :: Typeable a
+  => Worksheet
+  -> PointYX
+  -> CellType a
+  -> Either ExperimentParsingError a
```
gromakovsky: It can be `ParserType a`, but maybe you intentionally wrote `Either` here, I am not sure what's better.

backend/src/Edna/ExperimentReader/Types.hs
gromakovsky: I see that you changed it to be `newtype` and that `Codec.Xlsx` apparently takes position as `(Int, Int)`, so it's sort of justified.

backend/src/Edna/ExperimentReader/Parser.hs
```diff
@@ -0,0 +1,123 @@
...
+      -- If signal value is in brackets then it is outlier
+      (readParen True reads . toString . replace "," "." -> ((d, "") : _))
+      )) -> Right $ Signal d True
+
+    (_, cv) -> Left $ UnexpectedCellType p cellType cv
+
+-- | Check that cell representation equals to some value
+cellSatisfy :: Worksheet -> Text -> PointYX -> Bool
+cellSatisfy workSheet value (PointYX position) = case workSheet ^? ixCell position . cellValue of
+  Just (Just x) -> pretty x == value
```
gromakovsky: I don't like it because it will work weirdly if a cell contains `CellRich` or `CellError` for example. AFAIU we always expect text here, so maybe let's do `== value` only when we encounter `CellText` and return `False` for all other `CellValue` constructors? Like, if we pass `Text`, we probably expect `CellText`. If we want to support other types of cells, we should pass `CellValue` here (or `Maybe CellValue`).

backend/src/Edna/ExperimentReader/Parser.hs
```diff
@@ -0,0 +1,123 @@
...
+      restParams <- constructParams lastId (id2 : ids)
+      pure $ Parameter name (id1, id2 - 1) : restParams
+
+parseExperimentXls :: L.ByteString -> ParserType [ExperimentalMeasurement]
+parseExperimentXls content = do
+  xlsx <- first FileParsingError $ toXlsxEither content
+  workSheet <- maybeToRight WorksheetNotFound $ xlsx ^? xlSheets . ix 0 . _2
+  processWorkSheet workSheet
+
+processWorkSheet :: Worksheet -> ParserType [ExperimentalMeasurement]
```
gromakovsky, line 93: We should probably have nested maps or something like that and group measurements by projects, targets… Then the size of data will be much smaller and we won't have that many layers in list monad. But the current approach is ok for now, let's revise it later.

backend/src/Edna/ExperimentReader/Parser.hs
```diff
@@ -0,0 +1,123 @@
...
+    constructParams _ [] = pure []
+    constructParams lastId [idx] = do
+      name <- specificCellAt workSheet (direction (normalAxisPoint, idx)) CText
+      pure [Parameter name (idx, lastId)]
+    constructParams lastId (id1 : id2 : ids) = do
+      name <- specificCellAt workSheet (direction (normalAxisPoint, id1)) CText
+      restParams <- constructParams lastId (id2 : ids)
+      pure $ Parameter name (id1, id2 - 1) : restParams
+
+parseExperimentXls :: L.ByteString -> ParserType [ExperimentalMeasurement]
```
gromakovsky, line 87: It's an important top-level function, so please add a comment to it, even if its semantics is quite intuitive.

backend/src/Edna/ExperimentReader/Types.hs
```diff
@@ -0,0 +1,61 @@
...
+  build CDouble = "double"
+  build CSignal = "signal"
+
+instance Show (CellType a) where
+  show = pretty
+
+instance Eq (CellType a) where
+  (==) a b = (pretty a :: Text) == pretty b
+
+instance Buildable CellValue where
```
gromakovsky, line 59: Note that it raises the following warning:
```
[ 5 of 12] Compiling Edna.ExperimentReader.Types ( src/Edna/ExperimentReader/Types.hs, /home/gromak/Work/biocad/edna/backend/dist-newstyle/build/x86_64-linux/ghc-8.10.3/edna-0/noopt/build/Edna/ExperimentReader/Types.dyn_o )

src/Edna/ExperimentReader/Types.hs:36:1: warning: [-Worphans]
    Orphan instance: instance Buildable CellValue
    To avoid this
        move the instance declaration to the module of the class or of the type, or
        wrap the type with a newtype and declare the instance on the new type.
   |
36 | instance Buildable CellValue where
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

If this instance is needed only for `cellSatisfy`, it may be better to simply remove it, see my comment around that function. That being said, I am more or less ok with having this instance, so if you want to keep it, let's keep it and just suppress `-Worphans` in this particular file.

backend/src/Edna/ExperimentReader/Parser.hs
sashasashasasha151: >we always expect text here

No, it checks that cell text representation is the same as we want

backend/src/Edna/ExperimentReader/Parser.hs
sashasashasasha151: What do you mean by `nested maps`? \
Yes, we need to group them but I think we don't have understanding how we want to do it

backend/src/Edna/ExperimentReader/Types.hs
sashasashasasha151: Not only for `cellSatisfy`, also for error showing

backend/test/Test/Parser.hs
sashasashasasha151: What about path in cabal file?

backend/src/Edna/ExperimentReader/Parser.hs
gromakovsky: I mean, this function should always fail when it encounters cell type other than `CellText`, no? Because we pass only empty string and `<>` IIRC.

backend/src/Edna/ExperimentReader/Parser.hs
gromakovsky: Maybe `Map Target (Map CompoundId [ExperimentalMeasurement])` or something like.

>  Yes, we need to group them but I think we don't have understanding how we want to do it

Yes, I don't know what's the best way to represent it, that's why the current approach is ok for now.

backend/test/Test/Parser.hs
gromakovsky: Oh, interesting, I actually don't know and couldn't quickly find the answer. Probably the best way to know it is to test it on Windows ourselves… I would not bother about that for now, just make sure that paths are platform-independent in `.hs` files.

backend/src/Edna/ExperimentReader/Parser.hs
sashasashasasha151: No, why?\
Oh, wait, I now understand what you mean. You mean that I am using it only for comparing `CellText` values, right? That make sense, but initially it was made to compare for example not only `CellText` values but also `"1,1" == CellDouble 1,1`\
So, if you want I can remove it

backend/test/Test/Parser.hs
gromakovsky: We've decided to temporarily use `tasty-hspec`. Tomorrow I will either replace all `hspec` usages with `tasty` myself or throw away `tasty` instead.

backend/src/Edna/ExperimentReader/Parser.hs
gromakovsky: > You mean that I am using it only for comparing `CellText` values, right?

Yes.

> but initially it was made to compare for example not only `CellText` values but also `"1,1" == CellDouble 1,1`

Yes, and I don't like it because it complicates semantic of this function because now one needs to think whether `CellText "1,1"` is treated the same way as `CellDouble 1.1` So if you want it to be general, I think it's better to pass `CellValue` rather than just `Text`. But in this case it doesn't need to be general, so I would simply restrict it to `CellText` values.

gromakovsky (approved):
Ok, I think it's good enough, pls don't forget to cleanup the commit history.

PR #15, opened 2021-03-10
[EDNA-27] Update DB schema

gromakovsky:
## Description

Here I update DB schema (`initial_schema.sql`) according to changes in the product specification.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-27

- Tests
  - [ ] If I added new functionality, I added tests covering it. ← no because currently we don't have DB integration yet
        silently reappearing again.

- Documentation


gromakovsky:
I'll probably rename `initial_schema.sql` to `init.sql` in this PR. Let me know if you have any objections.

backend/sql/drop.sql
```diff
@@ -1,7 +1,10 @@
 drop table if exists project cascade;
```
pva701, line 1: perhaps it's good pr to change `drop.sql` to `drop schema public` :putya:

pva701 (approved):
good

backend/sql/drop.sql
gromakovsky: That's another issue ([EDNA-32](https://issues.serokell.io/issue/EDNA-32)), maybe someone will do it one day :agaaga:

backend/sql/init.sql
```diff
@@ -0,0 +1,172 @@
...
+    name            text        not null unique,
+    creation_date   timestamp   not null
+);
+
+create table if not exists compound
+(
+    compound_id     serial          not null primary key,
+    name            text            not null unique,
+    creation_date   timestamp       not null,
+    chemsoft_link   varchar(1000)   null
```
sashasashasasha151, line 38: What is chemsoft?

backend/sql/init.sql
```diff
@@ -0,0 +1,172 @@
...
+    project_id         int       not null,
+    methodology_id     int       null,  -- normally not null, but methodology can be removed
+    upload_date        timestamp not null,
+    -- Arbitrary metadata from the file
+    meta               json      null,
+
+    -- Name of the file
+    name               text      not null,
+    -- .xlsx blob
+    contents           bytea     not null,
```
sashasashasasha151, line 58: I do not know but is it a right way to store files?

backend/sql/init.sql
```diff
@@ -0,0 +1,172 @@
...
+            -- Deletion is currently not supported for compounds
+            on delete no action,
+
+    constraint has_target
+        foreign key (target_id) references target (target_id)
+            -- Target id is immutable
+            on update no action
+            -- Deletion is currently not supported for targets
+            on delete no action
+);
```
sashasashasasha151: What about `has_experiment_file`?

backend/sql/init.sql
```diff
@@ -0,0 +1,172 @@
...
+            on delete no action
+            on update no action,
+
+    constraint such_experiment_exists
+        foreign key (experiment_id) references experiment (experiment_id)
+            on delete no action
+            on update no action
+
+    -- TODO do we want to make sure that there is at least one measurement for (experiment_id, compound_id) in
+    -- measurement table?
```
sashasashasasha151, line 163: If user can remove experiments then we do not need to think about it

backend/sql/init.sql
```diff
@@ -0,0 +1,172 @@
...
+            -- Deletion is currently not supported for targets
+            on delete no action
+);
+
+create table if not exists measurement
+(
+    measurement_id  serial          not null primary key,
+    experiment_id   int             not null,
+    concentration   double precision not null,
+    signal          double precision not null,
```
sashasashasasha151, line 117: Lacks of `outlier` field (points disabled by default)

backend/sql/init.sql
```diff
@@ -0,0 +1,172 @@
...
+-- for each concentration. Usually there are multiple repetitions,
+-- so more than 1 value for each concentration.
+create table if not exists experiment
+(
+    experiment_id       serial  not null primary key,
+    -- File from which this experiment was taken.
+    -- Project, methodology and some other data are inherited from that file.
+    experiment_file_id  int     not null,
+    compound_id         int     not null,
+    target_id           int     not null,
```
sashasashasasha151, line 88: So, as I can't find smth useful in notion, will take design as a base\
In design we chose `experiment` by project, compound and target, but here I don't see `project_id`. So, if we will add file to this choice we should add `project_id` and corresponding constraints (by the way there is no `has_experiment_file` constraint).

sashasashasasha151 (changes requested):

backend/sql/init.sql
gromakovsky: It's some service that stores data about compounds AFAIK. We should display link to given compound in that service. It's optional and is null by default, but user can provide it for any compound and we will display it.

backend/sql/init.sql
gromakovsky: Hm, I was thinking a bit about it, but not too much. Since you asked, I tried to google a bit. It looks like a controversial topic without a single right answer. AFAIU it can be split into 2 questions:
1. Should we store files in DB or FS? We can store some _pointers_ in DB and store files as normal files of course.
2. PostgreSQL-specific question: if we decide to store in DB, should we use `bytea` or something else?

I think the primary criterion for the first question is the size of files we are going to store. The largest file we have now has 30270 bytes. I don't know how large files we should expect in practice, that's an open question, but probably we can set some reasonable limits ourselves saying "it's just MVP". So I think it's fine to say everything works fine on files less than 240kb, that's enough to store 8 times more data than the largest example we have. And probably average file will be ≈10-30kb.

Now the question is whether 10-30kb on average and 240kb in extreme case is a lot or not. In a [random old paper from 2006](https://www.microsoft.com/en-us/research/publication/to-blob-or-not-to-blob-large-object-storage-in-a-database-or-a-filesystem/?from=http%3A%2F%2Fresearch.microsoft.com%2Fapps%2Fpubs%2Fdefault.aspx%3Fid%3D64525) they say:
> As expected from the common wisdom, objects smaller than 256K are best stored in a database while objects larger than 1M are best stored in the filesystem. 

Of course it's old data and not about PostgreSQL and each case is unique and so on. But I think 256K is a reasonable size to distinguish "data that is not too big to be stored as blob in DB" and "data that is too big".

Ideally we should do some performance testing and make a decision based on that. At this stage I would just pick the simplest option and I _guess_ that storing files as blobs in DB is simpler, you can just access them directly from DB instead of accessing pointers and then reading from FS. And there is no need to think about consistency between files and pointers. There is one SO [comment](https://softwareengineering.stackexchange.com/a/150816) that I liked.

> I would be pragmatic about it, and follow the "don't optimize yet" principle.

> storing the data in the database does relieve you from dealing with other problems, e.g. where should the files be stored, how should they be backed up, etc.

> I would personally choose to store the data in the database, but make sure that the BLOBS are not read until they are really needed, i.e. no "SELECT * FROM ..." executed on those tables containing blogs.

All of that makes sense to me in our case. So, unless you have a good reason why you think that storing files differently will be _easier_, I would store them this way.

Regarding `bytea` specifically, I googled a bit and apparently there are two options provided by PostgreSQL:
1. `bytea`: https://www.postgresql.org/docs/9.0/datatype-binary.html. They say: "The SQL standard defines a different binary string type, called BLOB or BINARY LARGE OBJECT. The input format is different from bytea, but the provided functions and operators are mostly the same". Sounds appropriate in our case.
2. Streaming approach: https://www.postgresql.org/docs/current/largeobjects.html

I wouldn't bother with streaming prematurely, it's more like an optimization.

backend/sql/init.sql
gromakovsky: Makes sense, will add.

backend/sql/init.sql
gromakovsky: Currently removal will be defined only for test methodologies, so it won't be possible to remove experiments. It's an old comment and I didn't want to touch it because I don't know if we want such constraint :)

I mean, I can add a constraint checking that each experiment and sub-experiment has at least one point, but I am not sure it will be useful. I think it will be clearer when we implement relevant functionality, that's why I keep it here for now.

backend/sql/init.sql
gromakovsky: `project_id` is present in the `experiment_file` table and we refer to it. So if we want to filter by project, we can filter experiment files first and consider experiments only from those files.

backend/sql/init.sql
sashasashasasha151: Nice, thanks

backend/sql/init.sql
sashasashasasha151: But one project may have several experiment types, so with selectors on design we are not able to choose an experiment

gromakovsky: We don't need to choose one experiment, we need to filter all experiments by projects, compounds and targets. And each of these filters is optional. If someone selects project `proj0`, we will make a request which will fetch all experiments such that experiment file corresponding to this experiment has project equal to `proj0`. I didn't use SQL for a long time, but AFAIK it should be possible and not hard.

backend/sql/init.sql
gromakovsky: When a file is uploaded, we should probably create one sub-experiment for each experiment. If there are no outliers, we simply create a sub-experiment with no removed measurements. In presence of outliers, we can create a sub-experiment with remove measurements corresponding to outliers. So, I don't know if we should store information whether some measurement is an outlier, but it seems trivial to support and should be harmless, so I'll do it.

backend/sql/init.sql
sashasashasasha151: Please look at dashboard page, we have selecters here, what for?

gromakovsky: Selectors are filters, there are 3 filters. We show all experiments complying to these filters on the right.

gromakovsky: Here all 3 filters are used and there is a number of experiments on the right.

image

sashasashasasha151 (approved):

PR #19, opened 2021-03-12
[EDNA-40] Add new endpoints

gromakovsky:
## Description

Problem: frontend needs to be able to do various things, but currently no endpoints are available except for the legacy file upload one.

Solution: add all endpoints that should be necessary to implement Upload and Library pages.
Some instances were moved to `Types.hs`, so that they are no longer orphans.
New types were created.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-40

- Tests
  - [x] If I added new functionality, I added tests covering it
        silently reappearing again.

- Documentation


backend/src/Edna/Web/Types.hs
```diff
@@ -17,4 +43,180 @@ data ExperimentalMeasurement = ExperimentalMeasurement
...
+  { unFileSummary :: [FileSummaryItem]
+  } deriving stock (Generic, Show)
+
+-- | One element in 'FileSummary'. Corresponds to one target from the file.
+-- Contains all compounds that interact with the target in the file.
+-- Also contains information whether this target is new or already known.
+data FileSummaryItem = FileSummaryItem
+  { fsiTarget :: SqlId Target
+  -- ID of a target from the file.
+  , fsiCompounds :: [SqlId Compound]
```
gromakovsky: @pva701 this is what I return from `parseFile`. The problem is that when the file is parsed, it's not saved to DB and some compounds and targets can be new. So they have no ID. So I can't return IDs in this case. Is it ok if I return names in both `fsiTarget` and `fsiCompounds`? Is it ok if I do the same in `uploadFile` or do you want `FileSummary` to be different there? Or maybe we don't need `FileSummary` in the output of `uploadFile` at all.

backend/src/Edna/ExperimentReader/Parser.hs
```diff
@@ -13,7 +13,7 @@ import qualified Data.Map as M
 import Codec.Xlsx (CellValue(..), Worksheet(..), cellValue, ixCell, toXlsxEither, wsCells, xlSheets)
 import Data.List.Split (wordsBy)
 import Data.Text (replace)
-import Lens.Micro (ix, _Just)
+import Lens.Micro (_Just, ix)
```
sashasashasasha151, line 16: stylish haskell fix say that it is wrong 🤔

backend/src/Edna/Web/API.hs
```diff
@@ -1,29 +1,41 @@
...
-  { -- | Upload one experiment
+data EdnaEndpoints route = EdnaEndpoints
+  { -- | Legacy: upload one experiment
     eeUploadExperiment :: route
       :- "experiment"
       :> Summary "Upload an EXCEL file describing one experiment"
       :> MultipartForm Mem (MultipartData Mem)
       :> Post '[JSON] [ExperimentalMeasurement]
+
+  , eeFileUploadEndpoints :: route :- FileUploadAPI
```
sashasashasasha151: You can add `Tag` here from [servant util](https://github.com/serokell/servant-util/blob/master/servant-util/src/Servant/Util/Combinators/Tag.hs)

backend/src/Edna/Web/API.hs
```diff
@@ -32,3 +44,159 @@ type EdnaAPI =
```
sashasashasasha151, line 44: What for do we need `api` on the previous line?

backend/src/Edna/Web/API.hs
```diff
@@ -32,3 +44,159 @@ type EdnaAPI =
 
 ednaAPI :: Proxy EdnaAPI
 ednaAPI = Proxy
+
+-- | Endpoints necessary to implement file uploading.
+data FileUploadEndpoints route = FileUploadEndpoints
+  { -- | Parse the file and return its summary for preview.
+    fueParseFile :: route
+      :- "file"
```
sashasashasasha151: No need to write it in each entry point, just write it in `EdnaEndpoints`

backend/src/Edna/Web/Types.hs
```diff
@@ -17,4 +43,180 @@ data ExperimentalMeasurement = ExperimentalMeasurement
...
-deriveJSON ednaAesonWebOptions ''ExperimentalMeasurement
+----------------
+-- General types
+----------------
+
+-- | A simple wrapper over 'Word'. At the data layer, we identify all entities
+-- with numeric IDs and this data type corresponds to such an ID.
+-- It has a phantom parameter type which the type of the object
+-- identified by this ID.
+newtype SqlId t = SqlId
```
sashasashasasha151, line 54: What are the benefits of adding phantom parameter here?

backend/src/Edna/Web/API.hs
```diff
@@ -32,3 +44,159 @@ type EdnaAPI =
...
+      :> QueryParam "sortby" StubSortBy
+      :> Get '[JSON] [WithExtra Project ProjectExtra]
+
+  , -- | Get project data by ID
+    peGetProject :: route
+      :- "project"
+      :> Summary "Get project data by ID"
+      :> Capture "projectId" (SqlId Project)
+      :> Get '[JSON] (WithExtra Project ProjectExtra)
+  } deriving stock (Generic)
```
sashasashasasha151: Will write here comments which which also relate to the code below.\

```haskell
:> ReqBody '[JSON] Project
:> Post '[JSON] (WithId Project)
```
Why you do not write `WithExtra` here? If you add project from `Library` page you need to know its extra information\

```haskell
      :> Put '[JSON] (WithId Project)
```
Same here, don't understand why we do not want to get extra information\

You can combine edit and add entry-points making id optional\



```haskell
WithExtra Project ProjectExtra
```
So, as for me `WithExtra` and `WithId` are not needed. As we want to use MVC model, I propose to just create data types like `ProjectUpdateReq` and `SqlId Project` for requests (also I am not sure that `SqlId` is a good naming for web entry points) and `ProjectRep` for responses

sashasashasasha151 (changes requested):
We have enough entry-points to split them by different folders. I do not ask you to do it here, I want to do it in a separate pr after creating db schema. So, just need you approval for it

backend/src/Edna/Web/Types.hs
```diff
@@ -17,4 +43,180 @@ data ExperimentalMeasurement = ExperimentalMeasurement
...
+    "name" -> pure SortByName
+    "something" -> pure SortBySomething
+    x -> Left $ "unknown sorting order: " <> x
+
+----------------
+-- Entities
+----------------
+
+-- | Input data submitted along with uploaded file.
+data FileUploadInput = FileUploadInput
```
sashasashasasha151: Maybe FileUploadReq?

backend/src/Edna/Web/Types.hs
```diff
@@ -17,4 +43,180 @@ data ExperimentalMeasurement = ExperimentalMeasurement
...
+-- Contains all compounds that interact with the target in the file.
+-- Also contains information whether this target is new or already known.
+data FileSummaryItem = FileSummaryItem
+  { fsiTarget :: SqlId Target
+  -- ID of a target from the file.
+  , fsiCompounds :: [SqlId Compound]
+  -- IDs of all compounds interacting with this target.
+  , fsiIsNew :: Bool
+  -- Whether this target is new, i. e. wasn't uploaded before.
+  } deriving stock (Generic, Show)
```
sashasashasasha151: Can you please explain the logic of this params and corresponding entrypoints (it is duplicate of the comment above, but I want to explain my opinion). First of all what for we need to separate `parse` and `upload` entry points. Secondly I don't understand why here we do not return points and return only information about compounds and targets.

backend/src/Edna/Web/Types.hs
```diff
@@ -17,4 +43,180 @@ data ExperimentalMeasurement = ExperimentalMeasurement
...
+
+----------------
+-- Entities
+----------------
+
+-- | Input data submitted along with uploaded file.
+data FileUploadInput = FileUploadInput
+  { fuiProject :: SqlId Project
+  -- ^ ID of the project the file belongs to.
+  , fuiTestMethodology :: SqlId TestMethodology
```
sashasashasasha151: We also have description on the design.

backend/src/Edna/Web/Types.hs
```diff
@@ -17,4 +43,180 @@ data ExperimentalMeasurement = ExperimentalMeasurement
...
+  } deriving stock (Generic, Show)
+
+-- | Compounds are not submitted directly by users, so for now
+-- there is only one representation for frontend.
+-- MDe links are trivial to generate, so we offload this task to frontend.
+data Compound = Compound
+  { cName :: Text
+  -- ^ Name of the compound, it may be changed to be a number later.
+  , cChemSoft :: Maybe URI
+  -- ^ Link to ChemSoft.
```
sashasashasasha151, line 148: No `CreationDate`

sashasashasasha151 (changes requested):

backend/src/Edna/Web/Types.hs
sashasashasasha151: And there is also a lack of description in DB schema

backend/src/Edna/Web/Types.hs
sashasashasasha151: Okay, now I understand why we separate them, but question about points is still interests me

backend/src/Edna/Web/Types.hs
pva701: yeah, the reasonable point. I think for now it's ok to return names from both endpoints. But I am not sure it's 100% final, as we are going to implement View button after file was parsed, maybe it will be better to return ID for the second endpoint to make it possible

gromakovsky: Ok, then maybe I'll return different types because AFAIU the current `FileSummaryItem` is good enough for `upload`, so I can just add a new type instead of _replacing_ this type.

pva701 (approved):
@gromakovsky I see you are having fun here programming for a long time, so I am not going to interrupt you :putya:

The only two minor things: it's a little bit quirky to have this hierarchy structure of id, item and extra. It's not something crucial of course but I would enjoy the all field lying in a flat structure, something like
```
{
  projectId: ..
  projectName: ..
  createdAt: ...
}
```
generics could rescue here. 

Also the second one: I am not sure that the client asked us to have a list of compounds within a project and moreover, list of projects containing the target for a target. 
As I understand it came from the design, and if list of compounds within a project seems more or less reasonable (though representation on UI doesn't seem to be realistically convenient, as it might be plenty of them ~ 10 but there is a list), but list of projects for a target is less obviously useful to me and moreover I can't estimate the realistic number of projects targeting some specific target.

backend/src/Edna/ExperimentReader/Parser.hs
gromakovsky: If you run `stylish-haskell -i src/Edna/ExperimentReader/Parser.hs` from the `backend/` folder does it change this line? If yes, what's your `stylish-haskell` version?

backend/src/Edna/Web/Types.hs
pva701: (just randomly saw this comment)

it's not needed here, compounds are not something what is created manually but automatically when a new file uploaded and our system is not a source of truth, so it can't be creation date but rather date when it was automatically added to our system.

backend/src/Edna/Web/API.hs
```diff
@@ -32,3 +44,159 @@ type EdnaAPI =
...
+-- but let's do it later.
+
+-- | Endpoints related to projects.
+data ProjectEndpoints route = ProjectEndpoints
+  { -- | Add a new project.
+    peAddProject :: route
+      :- "project"
+      :> Summary "Add a new project"
+      :> ReqBody '[JSON] Project
+      :> Post '[JSON] (WithId Project)
```
pva701: yes, as Sasha pointed out extra information would be useful here, we need to show it on frontend as it's added

backend/src/Edna/Web/API.hs
gromakovsky: Yes, but updating `servant-util` to support latest versions of dependencies is a non-trivial task, I couldn't do it quickly, so no `servant-util` for now.

backend/src/Edna/Web/API.hs
```diff
@@ -32,3 +44,159 @@ type EdnaAPI =
...
+
+  , -- | Update an existing project.
+    peEditProject :: route
+      :- "project"
+      :> Summary "Update an existing project"
+      :> Capture "projectId" (SqlId Project)
+      :> ReqBody '[JSON] Project
+      :> Put '[JSON] (WithId Project)
```
pva701: and extra here is also useful

backend/src/Edna/Web/API.hs
pva701: it's not worth efforts to compile servant util for it :putya:

pva701:
Also, it's really hard to envisage all needed data on endpoints so maybe when I start attaching ~a d to an ant~ frontend to backend I will realise that something is missing

backend/src/Edna/Web/API.hs
gromakovsky: I have no idea, it was added by @pva701 and AFAIK you already tried to get information from him about this `api`. Maybe he imagines that something will be added outside of `/api` :shrug: 

Anyway, I don't think it really matters, especially in the scope of this PR where I didn't change it.

backend/src/Edna/Web/API.hs
gromakovsky: :pogchamp:

backend/src/Edna/Web/Types.hs
gromakovsky: We will probably have some functions that take IDs as arguments and this way it will be easier to understand whose ID we are expecting. For example, if a function takes more than one ID argument, this phantom parameter will be very useful.

That said, currently this type resides at `Web` level and thus is not supposed to be used by business logic layer (which is below `Web`). So maybe we will move it somewhere below and will use actively. If we don't do it and keep it as `Web`-only type, maybe this phantom parameter will become useless and we'll delete it (which should be quite easy I suppose, just deletion).

These types are somewhat experimental and I imagine that something may change on Haskell side, my main goal was to keep API more or less stable from the frontend side and there it won't make any difference if we keep this phantom parameter, it will still be a number.

backend/src/Edna/Web/API.hs
gromakovsky: > Why you do not write WithExtra here? If you add project from Library page you need to know its extra information\

Because this is what @pva701 [told me](https://serokell.slack.com/archives/C01F8AQSHGE/p1615480272097200?thread_ts=1615459098.060600&cid=C01F8AQSHGE). He is the consumer of this data, so I just return what he wants.

> You can combine edit and add entry-points making id optional\

Yes, I [was considering](https://serokell.slack.com/archives/C01F8AQSHGE/p1615461983072300?thread_ts=1615459098.060600&cid=C01F8AQSHGE) this. I don't think it matters, both options sound fine to me.

> So, as for me WithExtra and WithId are not needed. As we want to use MVC model, I propose to just create data types like ProjectUpdateReq and SqlId Project for requests (also I am not sure that SqlId is a good naming for web entry points) and ProjectRep for responses

I already like that I defined `WithExtra` and `WithId`, because now I can quickly change `WithId` to `WithExtra` and vice versa. Or remove ID and Extra completely.
I think `WithId` is useful at least for methodology and project endpoints because there were have `Project` and `TestMethodology` types as submitted by users and we need to return these types with `Id` attached. So instead of defining `ProjectRep` and `MethodologyRep` we can define one `WithId`. For `Compound` and `Target` it's less useful, but still I see a pattern here. Regarding `WithExtra` the situation is similar.

All in all I want to once again emphasize that these types may change in Haskell because currently we have no implementation and probably actual implementation will affect something. Maybe sometimes will be moved to lower levels, maybe not. Hopefully most of these changes will affect only backend, not frontend.

backend/src/Edna/Web/Types.hs
gromakovsky: I can rename if it's a more conventional naming.

backend/src/Edna/Web/Types.hs
gromakovsky: On the Upload page we have a preview table which summarizes the uploaded file. This is how it looks:
image

So `FileSummary` directly corresponds to that table. There is no place for points, it's a summary which lets one quickly check which compounds and targets will be added.

backend/src/Edna/Web/Types.hs
gromakovsky: I think the description in the design is useless, I don't understand why it's there, we never show it. I already [wrote](https://serokell.slack.com/archives/C01MH299XEU/p1615458841001300) about it a few days ago. I hope it will be removed soon, I'll remind about it.

backend/src/Edna/Web/Types.hs
gromakovsky: This type corresponds directly to this table from the wireframe:
image
 Currently there is no creation date there, so we don't return it. Though I see that we store it in DB, so we can return and let frontend decide what to do with it (in current design it can ignore this field).

Anyway, @pva701 and @sashasashasasha151, I want to point out that the set of tables in wireframes is really _tentative_, we just put some stuff that we find reasonable there, but something will likely change there when the customer starts testing our software. So for now let's just use our existing wireframe as the source of truth and implement everything according to it. AFAIU there is nothing very complex, but we can amend it if we heavily dislike something. I am also talking about @pva701's comment about showing projects for targets. We don't know what will should be shown for Targets, so we just picked some data that _probably_ makes sense. AFAIU we can get all projects by some SQL query, so it shouldn't be a big deal, though probably a bit bigger deal than returning some small data that's directly available.

Regarding `CreationDate`, I agree that there is inconsistency: I added it to `init.sql` because I thought that _maybe this information will be useful_, but at the same time I didn't add it here because it's currently not in the wireframe. I think I'll just add it to `Compound` to avoid inconsistency and frontend will ignore it for now.

Regarding naming, yes, compounds are not created manually, so it's a date of _automatic_ creation of an entry in our database.

backend/src/Edna/Web/API.hs
gromakovsky: You previously [told](https://serokell.slack.com/archives/C01F8AQSHGE/p1615480272097200?thread_ts=1615459098.060600&cid=C01F8AQSHGE) a different thing. But ok, I'll change it to `WithExtra`, it's a trivial change thanks to the current structure of types.

gromakovsky:
> The only two minor things: it's a little bit quirky to have this hierarchy structure of id, item and extra. It's not something crucial of course but I would enjoy the all field lying in a flat structure, something like

I can manually define aeson instances for `WithExtra` and `WithId`, but then the usefulness of these types will be smaller and probably we won't need them anymore :) But I'd like to keep them at this stage, because IMO they make API more uniform and make it easier to make changes, and because some types may be moved to lower levels.

So, if this thing is actually minor for you, I'd like to keep this structure for now, until we more or less stabilize backend. At least this way we'll probably merge this PR faster which is a good thing because a lot of future work depends on it.

> Also, it's really hard to envisage all needed data on endpoints so maybe when I start attaching a d to an ant frontend to backend I will realise that something is missing

Fortunately, as soon as we merge this PR you'll be able to update backend code in your frontend PRs whenever you need it.

backend/src/Edna/Web/Types.hs
gromakovsky: Also we may have types like `ProjectId`, `CompoundId`, etc. That's another feasible option, I think it will be clearer which one is better after we implement substantial part of backend.

backend/src/Edna/Web/API.hs
gromakovsky: > also I am not sure that `SqlId` is a good naming for web entry points

Oh, regarding name, there is again a question of where this type will reside eventually. And also I just couldn't find a good name, I could call it `Id`, but it's too general, so I needed some prefix and couldn't find a better one.

gromakovsky:
> We have enough entry-points to split them by different folders. I do not ask you to do it here, I want to do it in a separate pr after creating db schema. So, just need you approval for it

Probably your next task will be [EDNA-42](https://issues.serokell.io/issue/EDNA-42), so how about you do it only for Project and Methodology endpoints there (basically move `ProjectEndpoints` to `Edna.Project.API` or something like that, etc.) and then if we like it, we do the same for other endpoints in EDNA-43?

backend/src/Edna/Web/API.hs
gromakovsky: Actually one of the reasons why there is no `ProjectRep` is that response is different in different endpoints. In `getProject` we want to return all data (including extra), in `addProject` we just want to return `WithId Project`. It was the case on Thursday when @pva701 told me that. Now he apparently has changed his mind, so the structure of responses becomes more uniform. So it partially defeats the purpose of `WithExtra` at least.

To sum up, while I admit that these types are controversial (they were supposed to be experimental from the beginning), I am not convinced that we should change them _now_. So I propose to merge this PR with these types and revise them later.

backend/src/Edna/ExperimentReader/Parser.hs
sashasashasasha151: Yes, it does. `0.11.0.0`

backend/src/Edna/Web/Types.hs
sashasashasasha151: >easier to understand whose ID we are expecting

Just for devs? Or it can help us in writing logic?

backend/src/Edna/Web/Types.hs
sashasashasasha151: Wait, and what is under the view button?

backend/src/Edna/Web/Types.hs
sashasashasasha151: >Anyway, @pva701 and @sashasashasasha151, I want to point out that the set of tables in wireframes is really tentative, we just put some stuff that we find reasonable there, but something will likely change there when the customer starts testing our software.

But targets have it

backend/src/Edna/Web/API.hs
```diff
@@ -32,3 +44,157 @@ type EdnaAPI =
...
+
+-- TODO: pagination and sorting are just stubs for now (everywhere).
+-- Most likely we will use @servant-util@ to implement them,
+-- but let's do it later.
+
+-- | Endpoints related to projects.
+data ProjectEndpoints route = ProjectEndpoints
+  { -- | Add a new project.
+    peAddProject :: route
+      :- "project"
```
sashasashasasha151, line 76: Move it to `EdnaEndpoints` as you made for `"file"` and also move `"methodology"`, `"compound"` and `"targets"`

sashasashasasha151 (changes requested):

backend/src/Edna/Web/API.hs
gromakovsky: It's not so easy because we have `project` and `projects` here, there 2 different getters. The same is true in other structures. So I can't move it easily. I can keep only `project` endpoints here and extract `projects`, but then we will have more fields in `EdnaEndpoints` which is not very good. So IMO it won't be an improvement, it will have one advantage and one disadvantage.

backend/src/Edna/Web/Types.hs
gromakovsky: `View` is activated only after the file is added. When you click it, you go to Dashboard.

gromakovsky: By "activated" I mean that you can click on it.

backend/src/Edna/ExperimentReader/Parser.hs
gromakovsky: Mine is `0.12.2.0`, probably they changed something :shrug: 0.12 was a major update because they switched from `haskell-src-exts` to `ghc-lib-parser`. So I think some subtle changes are expected. If it only affects corner cases like this one (when an identifier starts with `_`), I think it's minor and we don't need to care.

backend/src/Edna/Web/Types.hs
gromakovsky: I mean that if we have some `f :: SqlId Project -> SqlId Target -> IO ()`, this parameter is helpful because this type is safer than `f :: SqlId -> SqlId -> IO ()` where we can easily confuse 2 IDs (pass in wrong order).

backend/src/Edna/Web/Types.hs
gromakovsky: Yes, and I guess that's one of the reasons why I put `creationDate` for `compound` in SQL. Anyway, as you can see, I've added `cAdditionDate :: Word64` to `Compound`, so I assume the problem is solved, isn't it?

backend/src/Edna/Web/Types.hs
sashasashasasha151: Make sense, but I still do not like this naming, but for now it is ok)

sashasashasasha151 (approved):

PR #20, opened 2021-03-15
[EDNA-39] Improve tests

gromakovsky:
## Description

This PR does some refactoring around tests, see individual commits for details. Note that it's based on #19, so it also has commits from there. I hope we'll merge #19 soon.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-39

- Tests
        silently reappearing again.

- Documentation


gromakovsky:
@sashasashasasha151 please note that here I change the way to run tests on dev machine.
1. There is no `scripts/test.sh` anymore, so you should run `make test` instead.
2. Additional test arguments can be passed using `TEST_ARGUMENTS` env variable. Example: `TEST_ARGUMENTS='--dry-run --no-color --match \"Valid tables\"' make test`.
3. I removed code that does some magic with test arguments before passing them to `--test-arguments`. The example above works for me, all arguments are propagated properly, so I don't think we need that magic. If somebody finds a case when it's necessary, we can add it back, but I personally couldn't find such a case, even an option with a space works (like `Valid tables` above).

sashasashasasha151 (approved):

gromakovsky:
@sashasashasasha151 just checking: your approval implies that you tried running tests and it worked well, so you don't mind using `make test` for this purpose, right?

sashasashasasha151:
> @sashasashasasha151 just checking: your approval implies that you tried running tests and it worked well, so you don't mind using `make test` for this purpose, right?

I ran tests and checked code. Yes, `make test` is ok for me. Moreover I like it

PR #21, opened 2021-03-15
[EDNA-43] File upload endpoints

gromakovsky:
## Description

Initial work on implementation of endpoints. Started with `parseFile` because it _almost_ doesn't need to access DB. Implemented it assuming that DB is empty. `parseFile` currently doesn't do anything smart, so it's already covered by existing parser tests. We should add tests for it when we have some way(s) to interact with DB. I'd like to do it in a separate PR.

This PR updates the handler where we implement endpoints to be `Edna`, so you can access DB there.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-43

- Tests
  - [x] If I added new functionality, I added tests covering it
        silently reappearing again.

- Documentation

### DIFF
--- a/backend/src/Edna/Web/Error.hs
+++ b/backend/src/Edna/Web/Error.hs
@@ -20,9 +20,11 @@ import Edna.ExperimentReader.Error (ExperimentParsingError(..))
 data EdnaServerError
   = XlsxParingError ExperimentParsingError
   | NoExperimentFileError
+  | TooManyExperimentFilesError
   deriving stock (Show, Generic)
 
-instance Exception EdnaServerError
+instance Exception EdnaServerError where
+  displayException = pretty
 
 -- | Class of exceptions which can be transformed to @ServerError@
 class Exception e => ToServerError e where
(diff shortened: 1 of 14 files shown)


backend/src/Edna/Web/Error.hs
```diff
@@ -31,9 +33,15 @@ class Exception e => ToServerError e where
...
-    err400 { errBody = encodeUtf8 (pretty @EdnaServerError @Text er) }
-  toServerError er@NoExperimentFileError =
-    err400 { errBody = encodeUtf8 (pretty @EdnaServerError @Text er) }
+  toServerError err = case err of
+    XlsxParingError _ ->
+      err400 { errBody = prettyErr }
+    NoExperimentFileError ->
+      err400 { errBody = prettyErr }
+    TooManyExperimentFilesError ->
+      err400 { errBody = prettyErr }
```
sashasashasasha151: Maybe just
```suggestion
  toServerError err = err400 { errBody = prettyErr }
```

sashasashasasha151 (approved):

backend/src/Edna/Web/Error.hs
gromakovsky: :mindblown:

gromakovsky: Fixed in place, gonna merge once CI passes.

PR #24, opened 2021-03-17
Adts for sql tables

sashasashasasha151:
## Description

This PR defines data types necessary to interact with PostgreSQL DB via `beam` and adds simple tests to check validity of these types.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-41

- Tests
        silently reappearing again.

- Documentation


backend/src/Edna/DB/Schema.hs
```diff
@@ -1,5 +1,325 @@
...
+deriving stock instance Eq (PrimaryKey MeasurementT Identity)
+
+--------------------------
+-- Analysis Method
+--------------------------
+
+data AnalysisMethodT f = AnalysisMethodRec
+  { amAnalysisMethodId :: C f (SqlSerial Int)
+  , amDescription :: C (Nullable f) Text
+  , amParameters :: C (Nullable f) (PgJSON String) -- TODO add corresponding type
```
gromakovsky: This should probably be just `()` for now, we don't have any parameters yet.

backend/src/Edna/DB/Schema.hs
```diff
@@ -1,5 +1,325 @@
...
+--------------------------
+-- Sub-experiment
+--------------------------
+
+data SubExperimentT f = SubExperimentRec
+  { seSubExperimentId :: C f (SqlSerial Int)
+  , seAnalysisMethodId :: C f Int
+  , seExperimentId :: C f Int
+  , seIsSuspicious :: C f Bool
+  , seResult :: C f (PgJSON String) -- TODO add corresponding type
```
gromakovsky: I suppose it should be `Double` for now, just one IC50 value.

backend/src/Edna/DB/Schema.hs
```diff
@@ -1,5 +1,325 @@
...
+
+--------------------------
+-- Experiment File
+--------------------------
+
+data ExperimentFileT f = ExperimentFileRec
+  { efExperimentFileId :: C f (SqlSerial Int)
+  , efMethodologyId :: C (Nullable f) Int
+  , efUploadDate :: C f UTCTime
+  , efMeta :: C (Nullable f) (PgJSON String) -- TODO add corresponding type
```
gromakovsky: I'll define a type to replace this `String`.

gromakovsky (commented):
Looks good so far, waiting for tests to check that all is fine and see how it all can be used.

backend/src/Edna/DB/Schema.hs
gromakovsky: If we merge #21 first, it can be `FileMetadata` from that PR.

backend/src/Edna/DB/Schema.hs
```diff
@@ -1,5 +1,327 @@
...
+import Database.Beam.Postgres (PgJSON, Postgres)
+import Database.Beam.Schema
+  (Beamable, C, Database, DatabaseSettings, Table(..), TableEntity, defaultDbSettings)
+
+--------------------------
+-- Project
+--------------------------
+
+data ProjectT f = ProjectRec
+  { pProjectId :: C f (SqlSerial Int32)
```
gromakovsky: They say that its range is `1 to 2147483647`, so I think `Word32` is slightly more accurate here, even though `Int32` should work too.

backend/test/Test/DB/DBSpec.hs
```diff
@@ -0,0 +1,107 @@
...
+  (genAnalysisMethodRec, genCompoundRec, genExperimentFileRec, genExperimentRec, genMeasurementRec,
+  genProjectRec, genRemovedMeasurementsRec, genSubExperimentRec, genTargetRec,
+  genTestMethodologyRec)
+import Test.Setup (ednaTestMode, withContext)
+
+spec :: Spec
+spec = withContext $ describe "Validate schema" $ modifyMaxSuccess (const 20) $ do
+  let EdnaSchema {..} = ednaSchema
+  let insertWithConflict schema value = runInsert' $ Pg.insert schema value $
+        Pg.onConflict (Pg.conflictingFields primaryKey) Pg.onConflictUpdateAll
```
gromakovsky: Do we expect conflicts in tests?

backend/test/Test/DB/DBSpec.hs
```diff
@@ -0,0 +1,107 @@
...
+  MeasurementT(..), ProjectT(..), SubExperimentT(..), TargetT(..), TestMethodologyT(..),
+  ednaSchema)
+import Test.DB.Gen
+  (genAnalysisMethodRec, genCompoundRec, genExperimentFileRec, genExperimentRec, genMeasurementRec,
+  genProjectRec, genRemovedMeasurementsRec, genSubExperimentRec, genTargetRec,
+  genTestMethodologyRec)
+import Test.Setup (ednaTestMode, withContext)
+
+spec :: Spec
+spec = withContext $ describe "Validate schema" $ modifyMaxSuccess (const 20) $ do
```
gromakovsky: I think `min 20` is slightly better than `const 20`, because in this case one can at least decrease the number of test runs. So it is 20 by default, but if I want 5 for some reason, I can specify 5 and it will work.

backend/test/Test/DB/DBSpec.hs
```diff
@@ -0,0 +1,107 @@
...
+  let EdnaSchema {..} = ednaSchema
+  let insertWithConflict schema value = runInsert' $ Pg.insert schema value $
+        Pg.onConflict (Pg.conflictingFields primaryKey) Pg.onConflictUpdateAll
+  let commonCheck schema expected = do
+        actual <- lift $ runSelectReturningOne' $ select $ all_ schema
+        liftIO $ actual `shouldBe` Just expected
```
gromakovsky: The logic of tests is good, but there are some problems and I don't quite understand what's going on. I tried to debug it, but couldn't succeed yet. So let me explain.
I see that CI fails with the following error:

>            HUnitFailure (Just (SrcLoc {srcLocPackage = "main", srcLocModule = "Test.DB.DBSpec", srcLocFile = "test/Test/DB/DBSpec.hs", srcLocStartLine = 33, srcLocStartCol = 18, srcLocEndLine = 33, srcLocEndCol = 49})) (ExpectedButGot Nothing "Just (MeasurementRec {mMeasurementId = SqlSerial {unSerial = 1}, mExperimentId = 1, mConcentration = 73.06692048416038, mSignal = 24.241577931890433, mIsOutlier = False})" "Just (MeasurementRec {mMeasurementId = SqlSerial {unSerial = 1}, mExperimentId = 1, mConcentration = 73.0669204841604, mSignal = 24.2415779318904, mIsOutlier = False})")

For some reason `Double` values are read a bit imprecisely. I don't know why it happens, it works as expected on my machine. I didn't try to investigate this particular problem, I'm leaving it to you.

The problem that I see here is that the error message looks bad, it is raw `HUnitFailure`. Inside `MonadTest` you should use `===` from `Hedgehog`, it prints the error more nicely and even shows the diff. So you can write `actual === Just expected` here. But when I did it, I was surprised because I started getting a completely weird error:
```
           SqlError {sqlState = "23503", sqlExecStatus = FatalError, sqlErrorMsg = "insert or update on table \"experiment\" violates foreign key constraint \"has_experiment_file\"", sqlErrorDetail = "Key (experiment_file_id)=(1) is not present in table \"experiment_file\".", sqlErrorHint = ""}
```

FTR I did this to break one test:
```
  it "Measurement" $ \ctx -> ednaTestMode ctx $
    insertMeasurement >>= (\m -> commonCheck esMeasurement (m {mSignal = 200}))
```
and commented out all other tests.

I still don't understand why it happens. I tried debugging it, even changed `client_min_messages` that we pass to Postgres to see its logs, but I still don't understand.

One thing I noticed is that if I set `modifyMaxShrinks (const 1)` (or 2 or 3), I will get an error about mismatch between `mSignal` as expected. If max number of shrinks is 11, I consistently get that error. Shrinks are not very useful in these particular tests, so as a workaround we can do `modifyMaxShrinks (const 3)` for example. But I am really curious why it happens and maybe indicates that we are doing something wrong. I'll probably stop at this point, but maybe will make one more attempt later today. I'll be glad if you figure out why it happens, also maybe it happens only on machine :shrug: That said, I don't think we should spend too much time on it because we have a lot of more urgent work. But since you worked with Postgresql from Haskell more than me, maybe you'll figure it our quickly.

2 more points:
1. In `ednaTestMode` currently you init and reset schema without `bracket`. Perhaps it makes sense to use `bracket` there, because AFAIU we want to reset shema every time. Alternatively, we can run `resetSchema` _before_ `initSchema`, AFAIU it simply does nothing if schema is not initialized. Maybe it will also solve the problem above.
2. `Hedgehog` prints code around the place that fails. It's based on `HasCallStack` machinery. When I used `===` in `commonCheck` and caused it to fail, I got a very large error message with almost the whole body of `spec`. I think we don't that much code in error messages, and AFAIU it can be solved by explicitly defining the type of `commonCheck` and adding `HasCallStack` constraint to it.

gromakovsky: I even tried `TEST_ARGUMENTS='-j1' make test`, but it also didn't help.

backend/src/Edna/DB/Schema.hs
sashasashasasha151: For all ids or how?

gromakovsky: Yes, I meant all IDs. I don't insist on changing `Int32` to `Word32` everywhere, it's really minor, I just think it's slightly more correct/accurate. So it's up to you.

backend/dev-config.yaml
```diff
@@ -6,7 +6,7 @@ api:
 db:
   # Syntax: https://www.postgresql.org/docs/9.5/libpq-connect.html#LIBPQ-CONNSTRING
   # It should match parameters from `/deployment/docker-compose.yml`
-  conn-string: "host=localhost dbname=ednadb user=ednauser password=ednapass"
+  conn-string: "host=localhost dbname=edna user=postgres password=12345"
```
gromakovsky: But why?

gromakovsky:
I have checked your fixes and squashed them.

backend/src/Edna/DB/Schema.hs
gromakovsky: I've done it myself.

backend/test/Test/Setup.hs
```diff
@@ -49,4 +49,7 @@ withContext = around withContext'
             ecDb . dbConnString .~ connString
       withPostgresConn testConfig $ \connPool -> do
         let ctx = EdnaContext testConfig connPool
-        callback ctx <* resetDbSchema ctx
+        callback ctx
+
+ednaTestMode :: EdnaContext -> PropertyT Edna a -> PropertyT IO ()
+ednaTestMode ctx = void . hoist (\action -> runRIO ctx $ schemaInit *> action <* runPg resetSchema)
```
gromakovsky: Also what do you think about my suggestion to wrap it into `bracket` or do `resetSchema` before `schemaInit`?

Also for SMT we also need to do `schemaInit` and `resetSchema`, but there we can't use `ednaTestMode`, so it would be nice to extract `schemaInit *> action <* runPg resetSchema` into another function.

backend/src/Edna/DB/Schema.hs
```diff
@@ -1,39 +1,331 @@
...
+deriving stock instance Eq (PrimaryKey ExperimentFileT Identity)
+
+--------------------------
+-- Experiment
+--------------------------
+
+data ExperimentT f = ExperimentRec
+  { eExperimentId :: C f (SqlSerial Int32)
+  , eExperimentFileId :: C f Int32
+  , eCompoundId :: C f Int32
```
gromakovsky: image

Here they suggest using `PrimaryKey` to refer to primary keys of other tables. Shouldn't `eCompoundId` have type `PrimaryKey CompoundT f`? The same question applies to a bunch of other places in this module.

backend/src/Edna/DB/Schema.hs
```diff
@@ -1,39 +1,331 @@
...
+
+data ExperimentFileT f = ExperimentFileRec
+  { efExperimentFileId :: C f (SqlSerial Int32)
+  , efProjectId :: C f Int32
+  , efMethodologyId :: C (Nullable f) Int32
+  , efUploadDate :: C f LocalTime
+  , efMeta :: C (Nullable f) (PgJSON FileMetadata)
+  , efDescription :: C f Text
+  , efName :: C f Text
+  , efContents :: C f ByteString
```
gromakovsky:
```suggestion
  , efContents :: C f LByteString
```

Because that's what we get in `uploadFile`.

backend/dev-config.yaml
sashasashasasha151: Will remove. Pushed it accidentally

backend/src/Edna/DB/Schema.hs
sashasashasasha151: I can't remember all the things that made me stop doing this. But as I remember it doesn't give you any advantages and if foreign key consists of two or more params you also can't write `PrimaryKey` for it. So, in the past I decided that it is better to not think about it as you still able to join tables

backend/test/Test/Setup.hs
sashasashasasha151: What is the difference between `schemaInit` before `resetSchema` and `resetSchema` before `schemaInit`? And how `bracket` solve this issue?

backend/src/Edna/DB/Schema.hs
gromakovsky: Okay, I didn't use this stuff before, so I don't know whether it's useful or not. Let's come to this topic later if we think that `PrimaryKey` here is better for some reason.

backend/test/Test/Setup.hs
gromakovsky: Currently if `action` throws an uncaught exception, `resetSchema` won't run. But I suppose this exception may be caught by testing engine afterwards (I am not sure if all exceptions are caught, but at least things like `HUnitFailure` should be caught for sure) and then the next test will run, but schema won't be reset when we run it.

If we use `bracket`, `resetSchema` will always happen. Or we can move it before `schemaInit` and then `resetSchema` will always happen if `schemaInit` happens.

sashasashasasha151: I made it in a little bit different way

backend/test/Test/DB/DBSpec.hs
```diff
@@ -0,0 +1,109 @@
...
+
+spec :: Spec
+spec = withContext $
+  -- We (almost) disable shrinking because when there are many shrinks something
+  -- goes wrong and we often get unexpected errors from SQL, like this one:
+  -- > insert or update on table "experiment" violates foreign key constraint "has_experiment_file"
+  describe "Validate schema" $ modifyMaxSuccess (min 20) $ modifyMaxShrinks (const 3) $ do
+    let EdnaSchema {..} = ednaSchema
+    let insertWithConflict schema value = runInsert' $ insert schema value {-$
+          Pg.onConflict (Pg.conflictingFields primaryKey) Pg.onConflictUpdateAll-}
```
gromakovsky: So, if it's not needed anymore, please don't forget to delete it.

gromakovsky (approved):

PR #27, opened 2021-03-19
[EDNA-43] Implement Upload service fully

gromakovsky:
## Description

1. Implemented `compoundNameToId` and `targetNameToId` used in `parseFile`, so now it's fully implemented.
2. Implemented `uploadFile`.
3. Added `UploadFile` to state machine tests.

## Related issue(s)

Resolves https://issues.serokell.io/issue/EDNA-43

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -19,3 +19,6 @@ Then run `make test` to execute tests.
 Run `stack exec -- edna-server -c dev-config.yaml` or `cabal run -- edna-server -c dev-config.yaml` depending on the build tool you used.
 * The server endpoints will be available at `http://localhost:9000/api/*`.
 * Swagger docs will be available at `http://localhost:9000/docs/`.
+
+Set `EDNA_DEBUG_DB=1` to enable logging of all DB actions made.
+It works in tests as well.
--- a/backend/sql/init.sql
+++ b/backend/sql/init.sql
@@ -27,14 +27,14 @@ create table if not exists target
 (
     target_id       serial      not null primary key,
     name            text        not null unique,
-    creation_date   timestamp   not null
+    creation_date   timestamp   not null default now()
 );
 
 create table if not exists compound
 (
     compound_id     serial          not null primary key,
     name            text            not null unique,
-    creation_date   timestamp       not null,
+    creation_date   timestamp       not null default now(),
     chemsoft_link   varchar(1000)   null
 );
 
@@ -45,12 +45,12 @@ create table if not exists compound
 -- it should be possible to download these files.
 create table if not exists experiment_file
 (
-    experiment_file_id serial not null primary key,
+    experiment_file_id serial    not null primary key,
     project_id         int       not null,
     methodology_id     int       null,  -- normally not null, but methodology can be removed
-    upload_date        timestamp not null,
+    upload_date        timestamp not null default now(),
     -- Arbitrary metadata from the file
-    meta               json      null,
+    meta               json      not null,
     -- Description provided from UI, it's like additional metadata
     -- that we store in a separate string
     description        text      not null,
@@ -132,8 +132,11 @@ create table if not exists analysis_method
     analysis_method_id serial not null primary key,
     description        text   null,
     -- Analysis-specific parameters
-    parameters         json   null
+    parameters         json   not null
 );
+
+-- Currently we have only one hardcoded analysis method.
+insert into analysis_method values (1, 'IC50', '[]'::json) on conflict do nothing;
 
 -- Sub-experiment is an experiment with disabled points.
 -- An experiment is a sub-experiment of itself with 0 disabled points.
--- a/backend/src/Edna/Upload/Service.hs
+++ b/backend/src/Edna/Upload/Service.hs
@@ -3,19 +3,25 @@
 module Edna.Upload.Service
   ( parseFile
   , uploadFile
+  , UploadError (..)
 
   -- * Exported for tests
   , parseFile'
+  , uploadFile'
   ) where
 
 import Universum
 
+import Database.Beam.Query (all_, guard_, select, val_, (==.))
+import Fmt (Buildable(..), pretty)
 import Lens.Micro.Platform (at, (?~))
 
+import Edna.DB.Integration (runSelectReturningOne')
+import Edna.DB.Schema (CompoundT(..), EdnaSchema(..), TargetT(..), ednaSchema)
 import Edna.ExperimentReader.Parser (parseExperimentXls)
 import Edna.ExperimentReader.Types
 import Edna.Setup
-import Edna.Web.Types
+import Edna.Web.Types hiding (cName, tName)
 
 -- | Parse contents of an experiment data file and return as 'FileSummary'.
 -- Uses database to determine which targets are new.
@@ -63,12 +69,30 @@ measurementsToSummary =
         , fsiCompounds = compounds
         }
 
+data UploadError =
+    UEUnknownProject (SqlId Project)
+  | UEUnknownTestMethodology (SqlId TestMethodology)
+  deriving stock (Show, Eq)
+
+instance Buildable UploadError where
+  build = \case
+    UEUnknownProject i -> "Unknown project: " <> build i
+    UEUnknownTestMethodology i -> "Unknown test methodology: " <> build i
+
+instance Exception UploadError where
+  displayException = pretty
 
 -- | Parse an experiment data file and save it to DB.
 uploadFile ::
   SqlId Project -> SqlId TestMethodology -> Text -> Text -> LByteString ->
   Edna FileSummary
-uploadFile _proj _methodology _description _fileName content = do
-  _parsedFile <- either throwM pure (parseExperimentXls content)
+uploadFile proj methodology description fileName content = do
+  uploadFile' proj methodology description fileName =<<
+    either throwM pure (parseExperimentXls content)
+
+uploadFile' ::
+  SqlId Project -> SqlId TestMethodology -> Text -> Text -> FileContents ->
+  Edna FileSummary
+uploadFile' _proj _methodology _description _fileName fc = do
   -- Stub implementation
-  parseFile content
+  parseFile' fc
--- a/backend/src/Edna/Upload/Service.hs
+++ b/backend/src/Edna/Upload/Service.hs
@@ -30,12 +30,22 @@ parseFile content =
 parseFile' :: FileContents -> Edna FileSummary
 parseFile' = measurementsToSummary . fcMeasurements
 
--- Need DB access for that
 compoundNameToId :: Text -> Edna (Maybe (SqlId Compound))
-compoundNameToId _ = pure Nothing
+compoundNameToId compoundName =
+  fmap (fmap mkSqlId) . runSelectReturningOne' $ select $ do
+    compound <- all_ (esCompound ednaSchema)
+    guard_ (cName compound ==. val_ compoundName)
+    return $ cCompoundId compound
 
 targetNameToId :: Text -> Edna (Maybe (SqlId Target))
-targetNameToId _ = pure Nothing
+targetNameToId targetName =
+  fmap (fmap mkSqlId) . runSelectReturningOne' $ select $ do
+    target <- all_ (esTarget ednaSchema)
+    guard_ (tName target ==. val_ targetName)
+    return $ tTargetId target
+
+mkSqlId :: Integral x => x -> SqlId y
+mkSqlId = SqlId . fromIntegral
 
 measurementsToSummary :: HashMap Text TargetMeasurements -> Edna FileSummary
 measurementsToSummary =
--- a/backend/src/Edna/Upload/Service.hs
+++ b/backend/src/Edna/Upload/Service.hs
@@ -69,9 +69,9 @@ measurementsToSummary =
       HashMap Text FileSummaryItem -> (Text, TargetMeasurements) ->
       Edna $ HashMap Text FileSummaryItem
     step acc (targetName, TargetMeasurements targetMeasurements) = do
-      target <- maybeToLeft targetName <$> targetNameToId targetName
+      target <- NameAndId targetName <$> targetNameToId targetName
       compounds <- forM (keys targetMeasurements) $ \compoundName ->
-        maybeToLeft compoundName <$> compoundNameToId compoundName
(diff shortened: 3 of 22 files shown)


gromakovsky:
There is a problem with the current state machine tests. When a new target is added in `applyUploadFile`, we assign its ID as `max_id + 1`. But actual implementation may assign a different ID. When it does `INSERT INTO targets`, `target_id` will be incremented by 1 even in case of unsuccessful insert (that may if we insert a target that is already added). So we can easily have 2 targets with IDs 1 and 10 in the DB, but in our model we assume that all IDs are consecutive.
So instead of assigning IDs ourselves we need to use IDs provided by actual implementation. We can take them from the result of `UploadFile` command, but inside `applyUploadFile` this result is available as `Var Result v`. And apparently it's impossible to get anything from this `Var Result`, we can't get `Var PartOfResult v` for example. So the only thing we can do is to put this `Var` into `EdnaState`. I tried to do that, but it gave a lot of complexity and I decided to postpone it.

I've also added some unit tests. We can't spend too much time on SMT at this stage, let's try to improve them later.

backend/src/Edna/Upload/Service.hs
```diff
@@ -30,12 +43,22 @@ parseFile content =
 parseFile' :: FileContents -> Edna FileSummary
 parseFile' = measurementsToSummary . fcMeasurements
 
--- Need DB access for that
 compoundNameToId :: Text -> Edna (Maybe (SqlId Compound))
-compoundNameToId _ = pure Nothing
+compoundNameToId compoundName =
+  fmap (fmap mkSqlId) . runSelectReturningOne' $ select $ do
```
sashasashasasha151, line 49: Maybe use `<<$>>`?

backend/src/Edna/Upload/Service.hs
```diff
@@ -3,19 +3,32 @@
...
 import Lens.Micro.Platform (at, (?~))
 
+import Edna.DB.Integration (runInsert', runInsertReturningList', runSelectReturningOne', transact)
+import Edna.DB.Schema as DB
 import Edna.ExperimentReader.Parser (parseExperimentXls)
-import Edna.ExperimentReader.Types
+import Edna.ExperimentReader.Types as EReader
 import Edna.Setup
-import Edna.Web.Types
+import Edna.Web.Types hiding (cName, tName)
```
sashasashasasha151: Are implicit imports match Serokell codestyle?

backend/src/Edna/Upload/Service.hs
```diff
@@ -53,12 +76,147 @@ measurementsToSummary =
         , fsiCompounds = compounds
         }
 
+data UploadError =
+    UEUnknownProject (SqlId Project)
+  | UEUnknownTestMethodology (SqlId TestMethodology)
+  deriving stock (Show, Eq)
```
sashasashasasha151: I think Errors should be in Errors module

backend/README.md
```diff
@@ -19,3 +19,6 @@ Then run `make test` to execute tests.
 Run `stack exec -- edna-server -c dev-config.yaml` or `cabal run -- edna-server -c dev-config.yaml` depending on the build tool you used.
 * The server endpoints will be available at `http://localhost:9000/api/*`.
 * Swagger docs will be available at `http://localhost:9000/docs/`.
+
+Set `EDNA_DEBUG_DB=1` to enable logging of all DB actions made.
```
sashasashasasha151, line 23: Maybe put it to config?

backend/sql/init.sql
```diff
@@ -132,9 +132,11 @@ create table if not exists analysis_method
     analysis_method_id serial not null primary key,
     description        text   null,
     -- Analysis-specific parameters
-    parameters         json   null
+    parameters         json   not null
 );
 
+insert into analysis_method values (default, 'IC50', '[]'::json);
```
sashasashasasha151: In dev mode (when we run this schema on each start without reset) it will insert in every time. I think you can write simple function here, which will check if `analysis_method` is empty or not and insert value

backend/src/Edna/Upload/Service.hs
```diff
@@ -53,12 +76,147 @@ measurementsToSummary =
...
+    targetNameToId targetName
+
+insertCompound :: Text -> Edna (Text, Word32)
+insertCompound compoundName = (compoundName,) . fromIntegral . unSqlId <$> do
+  runInsert' $ Pg.insert
+    (esCompound ednaSchema)
+    (insertExpressions [CompoundRec default_ (val_ compoundName) default_ default_])
+    (Pg.onConflict (Pg.conflictingFields cName) Pg.onConflictDoNothing)
+  fromMaybe (error $ "added compound not found: " <> compoundName)  <$>
+    compoundNameToId compoundName
```
sashasashasasha151, line 145: You can use `runInsertReturningOne` just here without calling `targetNameToId`.

sashasashasasha151: It is also possible to just run `runInsertReturningList` with `onConflictDoNothing` and get all new names

backend/src/Edna/Upload/Service.hs
```diff
@@ -53,12 +76,147 @@ measurementsToSummary =
...
+
+insertExperimentFile ::
+  Word32 -> Word32 -> FileMetadata -> Text -> Text -> LByteString -> Edna Word32
+insertExperimentFile projId methodId meta descr fileName blob = do
+  unSerial . expectOneInsertion . map efExperimentFileId <$> runInsertReturningList'
```
sashasashasasha151: `runInsertReturningOne` looks more convenient

backend/src/Edna/Upload/Service.hs
```diff
@@ -53,12 +76,147 @@ measurementsToSummary =
...
+      }
+
+insertExperiment :: Word32 -> Word32 -> Word32 -> [Measurement] -> Edna ()
+insertExperiment experimentFileId compoundId targetId measurements = do
+  expId <-
+    unSerial . expectOneInsertion . map eExperimentId <$> runInsertReturningList'
+    (insert (esExperiment ednaSchema) $ insertExpressions [experimentRec])
+  subExpId <-
+    unSerial . expectOneInsertion . map seSubExperimentId <$> runInsertReturningList'
+    (insert (esSubExperiment ednaSchema) $ insertExpressions [subExperimentRec expId])
```
sashasashasasha151: Maybe `runInsertReturningOne`

backend/src/Edna/Upload/Service.hs
```diff
@@ -53,12 +76,147 @@ measurementsToSummary =
...
+        , mConcentration = val_ (EReader.mConcentration measurement)
+        , mSignal = val_ (EReader.mSignal measurement)
+        , mIsOutlier = val_ (EReader.mIsOutlier measurement)
+        }
+    measurementId <-
+      unSerial . expectOneInsertion . map mMeasurementId <$> runInsertReturningList'
+      (insert (esMeasurement ednaSchema) $ insertExpressions [measurementRec])
+    when (EReader.mIsOutlier measurement) $
+      runInsert' $ insert (esRemovedMeasurements ednaSchema) $ insertValues
+      [RemovedMeasurementsRec subExpId measurementId]
```
sashasashasasha151: I am not sure that I am right, but as I understand call sql `1` time is better than calling it `n` times. So, you can construct list of `measurementRec` and insert them using `runInsertReturningList` and get it's `ids` with `isOutliers`, filter on `isOutliers` and again construct a list of sub-experiments and simply call sql one time

backend/src/Edna/Upload/Service.hs
```diff
@@ -53,12 +76,147 @@ measurementsToSummary =
...
+      , seAnalysisMethodId = val_ theOnlyAnalysisMethodId
+      , seExperimentId = val_ experimentId
+      , seIsSuspicious = val_ False
+      , seResult = val_ (PgJSON 10)  -- stub value, will be computed later
+      }
+
+expectOneInsertion :: (HasCallStack, Show x) => [x] -> x
+expectOneInsertion = \case
+  [x] -> x
+  xs -> error $ "Expected to insert 1 item, but inserted: " <> show xs
```
sashasashasasha151: Why not `specific` errors?

sashasashasasha151 (changes requested):

backend/src/Edna/Upload/Service.hs
gromakovsky: Yes, implicit imports from the current project are permitted.

backend/README.md
gromakovsky: I have several reasons why it's done this way, but the main one is that I need it to also work in tests and it's hard (maybe impossible) to modify config in tests without recompiling code.

backend/src/Edna/Upload/Service.hs
gromakovsky: No, I can't use `runInsertReturningOne` because it returns nothing if this compound already exists. We [discussed](https://serokell.slack.com/archives/C01F8AQSHGE/p1616340241019200) it yesterday.

backend/README.md
gromakovsky: If you mean "put it to config AND keep it as env variable", it probably makes sense, but it's extra work which is not needed now, we can do it in [EDNA-44](https://issues.serokell.io/issue/EDNA-44) when we have time.

backend/src/Edna/Upload/Service.hs
gromakovsky: Interesting, I don't know why I did it this way, probably because there is no `runInsertReturningOne'` in `Edna.DB.Integration`, but I guess I can add it.

backend/src/Edna/Upload/Service.hs
gromakovsky: Ok, sounds doable. I suppose there will be 2 SQL calls instead of this `forM_`, not 1.

backend/src/Edna/Upload/Service.hs
gromakovsky: What is `specific`? Anyway, this function will likely be removed because you say there is `runInsertReturningOne`.

backend/src/Edna/Upload/Service.hs
sashasashasasha151: But what about `runInsertReturningList`?

backend/src/Edna/Upload/Service.hs
sashasashasasha151: So, create smth like `data DBError = `. There are also other plain errors in this module

backend/sql/init.sql
gromakovsky: Oh right, but I guess I don't need a function here, I can do `insert into analysis_method values (1, 'IC50', '[]'::json) on conflict do nothing`.

backend/src/Edna/Upload/Service.hs
gromakovsky: Apparently there is no `runInsertReturningOne` in `beam`. But I still can define it.

sashasashasasha151: There is `runInsertReturning`, isn't it what you want?

sashasashasasha151: Looks like it isn't (

backend/src/Edna/Upload/Service.hs
gromakovsky: It will return empty list if nothing is updated and I will need to do `SELECT` after that. I can do the following: `INSERT RETURNING`, check whether anything was returned, and if not, call `SELECT`. This way when we add a new target/compound we will execute one command instead of two. But there will be more code in this case and I don't think we'll ever notice this optimization, it's very minor.

backend/src/Edna/Upload/Service.hs
gromakovsky: It looks a little bit different. Anyway, I've defined
```
-- | Insert items and expect one item in the result. The caller is responsible
-- for ensuring that exactly one result will be returned.
runInsertReturningOne' ::
  ( Beamable table
  , FromBackendRow Postgres (table Identity)
  ) => SqlInsert Postgres table -> Edna (table Identity)
runInsertReturningOne' = fmap expectOneInsertion . runInsertReturningList'
  where
    expectOneInsertion :: HasCallStack => [x] -> x
    expectOneInsertion = \case
      [x] -> x
      xs -> error $
        "Expected to insert 1 item, but inserted: " <> show (length xs)
```

and it's probably sufficient for now

backend/src/Edna/Upload/Service.hs
gromakovsky: Because this is a programmer mistake which should never happen. I generally follow [these guidelines](https://github.com/input-output-hk/cardano-sl/blob/2a5b5769799fd220e2ff6a6b09a1832d8e6f06c0/docs/exceptions.md). `data DBError` is useful when there is a chance that this error will be handled somehow because you can turn it into some other format (e. g. HTTP response) or even translate to another language. But this error is a programmer mistake which we don't need to handle specially.

gromakovsky:
@sashasashasasha151 I've accidentally squashed 2 fixup commits, but you still can check them here:
* https://github.com/serokell/edna/commit/03f53786273b78a7b613e57550244db42b21bdfd
* https://github.com/serokell/edna/commit/1e121b0430512e24c1332e8bacbbcb2b3f772cf4

The last thing I need to do here is to update `FileSummary` to always contain names. I'll do it shortly.

gromakovsky:
@sashasashasasha151 you are welcome to re-review.

sashasashasasha151:
> @sashasashasasha151 you are welcome to re-review.

I have email notifications. So, you can just rerequest my review

backend/src/Edna/Upload/Service.hs
sashasashasasha151: Why don't you like `<<$>>`?

sashasashasasha151 (approved):

backend/src/Edna/Upload/Service.hs
gromakovsky: Oh, right. Well, I don't like it for a couple of reasons: it's non-standard (people who are not used to `Universum` don't know this operator) and I always forget whether it's `<<$>>` or `<$$>`. And here there is an additional reason: it can't be used as an operator in infix position, it can be used only as a function and I slightly dislike that as well. So I think `fmap (fmap …` is ok here.

PR #30, opened 2021-03-22
Add health check to edna-server

gromakovsky:
## Description

Problem: sometimes we just want to check whether `edna-server` and
not involve any application logic into it.

Solution: apparently a quite standard solution is to have a `/health`
GET endpoint which returns 204. That's what we do here using `servant`.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-54

- Tests
        silently reappearing again.

Changes only affect endpoints and currently we don't have infrastructure for this type of tests, we'll work on it later.

- Documentation

### DIFF
--- a/backend/src/Edna/Web/API.hs
+++ b/backend/src/Edna/Web/API.hs
@@ -9,8 +9,8 @@ module Edna.Web.API
 
 import Universum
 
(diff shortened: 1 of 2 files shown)


pva701 (approved):
Sorry for so long approve

PR #31, opened 2021-03-22
Implement Library endpoints

gromakovsky:
## Description

This PR implements all Library endpoints. API schema has changed a bit.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-42
https://issues.serokell.io/issue/EDNA-43

- Tests
        silently reappearing again.

Tests will be added in the next PR.

- Documentation

### DIFF
--- a/backend/src/Edna/Library/Service.hs
+++ b/backend/src/Edna/Library/Service.hs
@@ -7,6 +7,11 @@ module Edna.Library.Service
   , getCompound
   , getCompounds
   , editChemSoft
+  , getMethodology
+  , getMethodologies
+  , deleteMethodology
+  , addMethodology
+  , updateMethodology
   ) where
 
 import Universum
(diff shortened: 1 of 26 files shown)


backend/src/Edna/Library/Web/Schema.hs
```diff
@@ -0,0 +1,11 @@
+module Edna.Library.Web.Schema
```
gromakovsky: I don't see a benefit in this module. It's small and has an orphan instance. This instance can be moved to `Edna.Library.Web.Types` harmlessly. I could understand this module if we had 100 library web types, but we are going to have about 10-20 and each instance is going to be rather small and trivial, so I think it's better to keep them in the same module as type definitions. And currently there is inconsistency because JSON instances are defined in `Types` while swagger instances are here.

backend/src/Edna/Library/Service.hs
```diff
@@ -0,0 +1,29 @@
+-- | Implementation of target functionality
```
gromakovsky: Don't forget to change `target` to `library`.

backend/src/Edna/Library/Web/Schema.hs
```diff
@@ -0,0 +1,11 @@
...
+  () where
```
gromakovsky: Btw, if export list is empty, our style guide permits writing `()` on the same line as `module …` and I think it looks a bit neater. But that's up to you (and I propose to remove this module anyway).

backend/dev-config.yaml
```diff
@@ -6,7 +6,7 @@ api:
 db:
   # Syntax: https://www.postgresql.org/docs/9.5/libpq-connect.html#LIBPQ-CONNSTRING
   # It should match parameters from `/deployment/docker-compose.yml`
-  conn-string: "host=localhost dbname=ednadb user=ednauser password=ednapass"
+  conn-string: "host=localhost dbname=edna user=postgres password=12345"
```
gromakovsky: TODO: undo.

backend/src/Edna/Library/Web/Types.hs
```diff
@@ -0,0 +1,47 @@
...
+
+-- | Targets are not submitted directly by users, so for now
+-- there is only one representation for frontend.
+data TargetResp = TargetResp
+  { tName :: Text
+  -- ^ Name of the target.
+  , tProjects :: [Text]
+  -- ^ Names of all projects where this target is involved.
+  , tCreationDate :: LocalTime
+  -- ^ UNIX timestamp when the target was created.
```
gromakovsky: It's not "UNIX timestamp" anymore btw.

sashasashasasha151 (approved):

PR #37, opened 2021-03-25
Use basic version of plotly

pva701:
## Description

This makes a bundle smaller.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-

- Tests
        silently reappearing again.

- Documentation


gromakovsky:
Is it ok that you removed `frontend/yarn.lock`?

frontend/src/api/EdnaApi.ts
```diff
@@ -81,7 +81,7 @@ export default function EdnaApi(axios: AxiosInstance): EdnaApiInterface {
           // TODO remove this mock transformation
           const byCompound = groupCompounds(response.data);
           return Object.entries(byCompound).map(([cmpId, measurements]) => ({
-            target: "target",
+            target: "",
```
gromakovsky, line 84: Why?

frontend/package.json
```diff
@@ -79,7 +79,7 @@
   "scripts": {
     "clean": "rm -rf dist/ && rm -rf .cache/",
     "serve": "webpack serve --config webpack.dev.js",
-    "build": "webpack --config webpack.dev.js",
+    "build": "webpack --config webpack.prod.js",
```
gromakovsky, line 82: How/where is `frontend/package.json` used? Why do we use different configs for `build` and `serve`?

frontend/src/api/EdnaApi.ts
pva701: because it's just dummy thing, which I will remove after all. for now it's "" because compoundId from old upload method already contains target and I wanted to have proper chart title

frontend/package.json
pva701: build just assembles bundles, in production mode bundles are minimises. for serve it's just dev only configuration which runs dev server and proxy, so dev config doesn't affect production and it may contain anything. ideally production and dev config should contain essential parts common, so this part is in `webpack.common.js`

pva701:
>frontend/yarn.lock

OH SHIT, no

gromakovsky (approved):

PR #40, opened 2021-03-26
Update local deployment

gromakovsky:
## Description

1. Updated `docker-compose.yml` to include `frontend` and `backend`, mostly just copied the config provided by @mkaito.
2. Added `deployment/config.yaml` which slightly differs from `backend/dev-config.yaml` because for example `init.sql` is in a different location.
3. Updated documentation.

## Related issue(s)

Resolves https://issues.serokell.io/issue/EDNA-50

- Tests
        silently reappearing again.

No tests because it's something too high-level.

- Documentation

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -1,6 +1,6 @@
 # Edna server
 
-Web server for Edna.
+HTTP API backend service for Edna.
 
 ## Build instructions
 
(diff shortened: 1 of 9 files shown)


backend/src/Edna/DB/Connection.hs
```diff
@@ -36,11 +38,23 @@ createConnPool
...
+  close        -- connection destroy action
+  1            -- number of individual pools (just one is fine)
+  nominalDay   -- maximum time the connection should remain open while unused
+  maxConnsNum  -- maximum number of DB connections.
+  where
+    aSecond = 1e6
+    -- We try to reconnect in case of an error because if we launch postgres
+    -- and edna-server at nearly the same time, edna-server may try to connect
+    -- before postgres starts accepting connections.
+    -- It may happen if you deploy using @docker-compose@ for example.
```
worm2fed, line 51: As I can see we use `depends_on` in docker-compose, so backend service going to be run after postgres one.
Anyway I think it's good enough to have such a mechanism for DB connection

worm2fed (approved):
Everything seems fine from my side.
We can add more env variables for docker-compose, but it's partially done in my task, so we can continue later.

backend/src/Edna/DB/Connection.hs
zhenyavinogradov: `depends_on` ensures that backend container does not start before postgres one, but we still cannot be sure that postgres has finished initializing and is ready to accept connections when backend tries to connect to it, it may take some time for postgres to initialize.

worm2fed: @zhenyavinogradov thanks for information, I'll remember it

backend/README.md
```diff
@@ -1,6 +1,6 @@
 # Edna server
 
-Web server for Edna.
+Web server (backend) for Edna.
```
mkaito: Excuse me for being pedantic, but it's an API backend service, not a "web server".

deployment/docker-compose.yml
```diff
@@ -1,17 +1,38 @@
...
     ports:
-      - "5432:5432"
+      - 5432:5432
     environment:
       POSTGRES_USER: ednauser
       POSTGRES_PASSWORD: ednapass
       POSTGRES_DB: ednadb
     volumes:
-      - ./postgres:/var/lib/postgresql/data/
+      - edna-db-data:/var/lib/postgresql/data/
```
mkaito, line 32: While a named volume is "cleaner", I suppose, it also means you don't have very easy manual access to the data. You have to run, say, a busybox container and mount the volume to manipulate it. The reason I used a host path is so that you could easily reset or manipulate the database during development.

gromakovsky: I treat `postgres` folder as black box that I shouldn't touch and use `psql` to interact with DB. AFAIK @pva701 and @sashasashasasha151 do something similar (they use IDE plugin instead of plain `psql`, but it shouldn't matter). So if I want to reset DB, I can do `\i sql/drop.sql;` or run `edna-server` with `init-with-drop`. So I think it's not a problem.

backend/README.md
gromakovsky: Fixed.

deployment/docker-compose.yml
zhenyavinogradov: I'm not sure if named volumes are cleaner, I just noticed the problems I described in the commit message and wanted to help with it while you were not around, but if you think there's a better way, feel free to change it back, or change it to something else.

Dropping a named volume is simple with `docker-compose down -v` or `docker volume rm`, but inspecting or changing it is harder, yes.

backend/README.md
zhenyavinogradov: By web server people usually mean anything serving http requests, a backend service matches this definition too

PR #42, opened 2021-03-29
[EDNA-55] Define Dashboard API

gromakovsky:
## Description

Problem: we are going to implement Dashboard page, but currently there
is no backend API for it.

Solution: define API types and schema. Endpoints will be implemented
later, currently we use stubs with 501 HTTP code.

Note: currently we don't define any endpoints necessary to drop points
and create new sub-experiments. They will be added later.

image

## Related issue(s)

https://issues.serokell.io/issue/EDNA-55

- Tests
        silently reappearing again.

We don't have API tests for now.

- Documentation

### DIFF
--- a/backend/src/Edna/Dashboard/Web/API.hs
+++ b/backend/src/Edna/Dashboard/Web/API.hs
@@ -1,0 +1,99 @@
+-- | Dashboard-related part of API definition along with implementation.
+
+module Edna.Dashboard.Web.API
+  ( DashboardEndpoints (..)
+  , DashboardAPI
+  , dashboardEndpoints
+  ) where
+
+import Universum
+
+import Servant (ReqBody)
+import Servant.API (Capture, Delete, Get, JSON, NoContent, Post, Put, QueryParam, Summary, (:>))
+import Servant.API.Generic (AsApi, ToServant, (:-))
+import Servant.Server (err501)
+import Servant.Server.Generic (AsServerT, genericServerT)
+
+import Edna.Dashboard.Web.Types
+import Edna.Setup (Edna)
+import Edna.Util (CompoundId, IdType(..), ProjectId, SubExperimentId, TargetId)
+import Edna.Web.Types (StubSortBy, WithId)
+
+-- TODO: pagination and sorting are just stubs for now.
+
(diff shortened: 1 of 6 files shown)


backend/src/Edna/Dashboard/Web/API.hs
```diff
@@ -0,0 +1,99 @@
...
+      :> Get '[JSON] ExperimentsResp
+
+  , -- | Get sub-experiment's (meta-)data by ID.
+    deGetSubExperiment :: route
+      :- "subExperiment"
+      :> Summary "Get sub-experiment's (meta-)data by ID"
+      :> Capture "subExperimentId" SubExperimentId
+      :> Get '[JSON] (WithId 'SubExperimentId SubExperimentResp)
+
+  , -- | Get sub-experiment's (meta-)data by ID.
```
pva701: wrong comment

backend/src/Edna/Dashboard/Web/Types.hs
```diff
@@ -0,0 +1,82 @@
...
+import Edna.Util
+  (CompoundId, IdType(..), MethodologyId, ProjectId, SubExperimentId, TargetId, ednaAesonWebOptions,
+  gDeclareNamedSchema)
+import Edna.Web.Types (WithId)
+
+-- | Experiment as response from the server.
+data ExperimentsResp = ExperimentsResp
+  { erExperiments :: [WithId 'ExperimentId ExperimentResp]
+  , erMeanIC50 :: Maybe Double
+  -- ^ If compound and target and selected,
```
pva701: are selected maybe?

pva701 (approved):
Looks more or less sane, but it's hard to predict precisely what else will be needed. AFAIU, there are no creation methods (like to create a new subexperiment?).

gromakovsky:
@pva701 

> AFAIU, there are no creation methods (like to create a new subexperiment?).

Yes, that's written in the PR description:
> Note: currently we don't define any endpoints necessary to drop points and create new sub-experiments. They will be added later.

The reason is that I just follow our roadmap (that I invented myself), according to which point dropping will be implemented after we are done with Dashboard that doesn't have it.

PR #43, opened 2021-03-29
Upload module structure refactoring

sashasashasasha151:
## Related issue(s)

https://issues.serokell.io/issue/EDNA-60

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/backend/src/Edna/Library/DB/Query.hs
+++ b/backend/src/Edna/Library/DB/Query.hs
@@ -16,46 +16,57 @@ module Edna.Library.DB.Query
   , getProjectsWithCompounds
   , insertProject
   , updateProject
+  , insertTarget
+  , getTargetByName
+  , getCompoundByName
+  , insertCompound
   ) where
 
 import Universum
 
 import qualified Data.List as L
+import qualified Database.Beam.Postgres.Full as Pg
 
 import Database.Beam.Backend (SqlSerial(..))
 import Database.Beam.Postgres (pgNubBy_)
 import Database.Beam.Query
   (all_, asc_, cast_, default_, delete, filter_, guard_, insert, insertExpressions, int, just_,
-  leftJoin_, orderBy_, select, update, val_, (<-.), (==.))
+  leftJoin_, lookup_, orderBy_, select, update, val_, (<-.), (==.))
 
 import Edna.DB.Integration
-  (runDelete', runInsertReturningOne', runSelectReturningList', runSelectReturningOne', runUpdate')
-import Edna.DB.Schema (EdnaSchema(..), ExperimentFileT(..), ExperimentT(..), ednaSchema)
+  (runDelete', runInsert', runInsertReturningOne', runSelectReturningList', runSelectReturningOne',
+  runUpdate')
+import Edna.DB.Schema (EdnaSchema(..), ExperimentT(..), ednaSchema)
 import Edna.Library.DB.Schema as LDB
-  (CompoundRec, CompoundT(..), ProjectRec, ProjectT(..), TargetRec, TargetT(..), TestMethodologyRec,
-  TestMethodologyT(..))
+  (CompoundRec, CompoundT(..), PrimaryKey(..), ProjectRec, ProjectT(..), TargetRec, TargetT(..),
+  TestMethodologyRec, TestMethodologyT(..))
 import Edna.Library.Web.Types
   (MethodologyReqResp(..), ProjectReq(..), ProjectResp(..), TargetResp(..))
 import Edna.Setup (Edna)
-import Edna.Util (IdType(..), SqlId(..), TargetId)
+import Edna.Upload.DB.Schema (ExperimentFileT(..))
+import Edna.Util as U (CompoundId, IdType(..), ProjectId, SqlId(..), TargetId, justOrError)
 import Edna.Util.URI (renderURI)
 import Edna.Web.Types (WithId(..))
 
-targetToDomain :: TargetId -> TargetRec -> [Maybe ProjectRec] -> WithId 'TargetId TargetResp
+--------------------------
+-- Target
+--------------------------
+
+targetToDomain :: TargetId -> TargetRec -> [Maybe ProjectRec] -> WithId 'U.TargetId TargetResp
 targetToDomain targetSqlId TargetRec{..} projects = WithId targetSqlId $ TargetResp
   { trName = tName
   , trProjects = mapMaybe (fmap pName) projects
   , trAdditionDate = tAdditionDate
   }
 
-getTargetById :: TargetId -> Edna (Maybe (WithId 'TargetId TargetResp))
+getTargetById :: TargetId -> Edna (Maybe (WithId 'U.TargetId TargetResp))
 getTargetById targetSqlId = do
   targetWithProjects <- targetsWithProjects $ Just targetSqlId
   case targetWithProjects of
     [] -> pure Nothing
     xs@((target, _) : _) -> pure $ Just $ targetToDomain targetSqlId target $ map snd xs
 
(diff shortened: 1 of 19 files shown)


backend/src/Edna/Library/DB/Query.hs
```diff
@@ -87,7 +98,25 @@ targetsWithProjects targetSqlId = runSelectReturningList' $ select $
...
+    (esTarget ednaSchema)
+    (insertExpressions [TargetRec default_ (val_ targetName) default_])
+    (Pg.onConflict (Pg.conflictingFields tName) Pg.onConflictDoNothing)
+  getTargetByName targetName
+
+--------------------------
+-- Compound
+--------------------------
+
+getCompoundById :: SqlId 'U.CompoundId -> Edna (Maybe CompoundRec)
```
gromakovsky:
```suggestion
getCompoundById :: CompoundId -> Edna (Maybe CompoundRec)
```

Perhaps it makes sense to replace all `SqlId 'FooId` to `FooId` in this file while we are at it.
Also I doubt you need explicit `U.` anywhere. `TargetId` constructor and `TargetId` type do not conflict with each other AFAIK (if that's the reason why you use `U.`).

backend/src/Edna/Library/DB/Query.hs
```diff
@@ -87,7 +98,25 @@ targetsWithProjects targetSqlId = runSelectReturningList' $ select $
...
       Nothing -> \_ -> val_ True
 
-getCompoundById :: SqlId 'CompoundId -> Edna (Maybe CompoundRec)
+getTargetByName :: Text -> Edna (Maybe TargetRec)
+getTargetByName name = runSelectReturningOne' $ select $ do
+  targets <- all_ $ esTarget ednaSchema
+  guard_ (LDB.tName targets ==. val_ name)
+  pure targets
+
+insertTarget :: Text -> Edna (Maybe (Text, Word32))
```
gromakovsky: 1. Why did you change it to return `Maybe`? It must always return `Just` and it's responsibility of _this_ function to return `Just`. The contract of this function is that it always inserts the target or returns ID of the existing one. This function knows that `getTargetByName` must return `Just` because it sets `onConflict` as necessary. The caller currently just assumes that this function always returns `Just` and uses `error` (aka "impossible happened"). If we remove `Maybe`, the caller won't need to think whether `Nothing` is possible.
2. Pairing ID with name sounds like something the DB layer doesn't need to care about. This layer doesn't know anything about API that we expose, and if we forget this knowledge it will be strange to return `(name, ID)` pair. Returning only ID is ok because it's auto-generated. Returning the whole object is also ok. But returning some combination of fields sounds like something the Service layer should take care of.

backend/src/Edna/Library/DB/Query.hs
```diff
@@ -96,16 +125,32 @@ getCompoundById (SqlId compoundId) = runSelectReturningOne' $ select $ do
...
+-- Test methodology
+--------------------------
+
 getMethodologyById :: SqlId 'MethodologyId -> Edna (Maybe TestMethodologyRec)
-getMethodologyById (SqlId methodologyId) = runSelectReturningOne' $ select $ do
-  methodologies <- all_ $ esTestMethodology ednaSchema
-  guard_ (tmTestMethodologyId methodologies ==. val_ (SqlSerial methodologyId))
-  pure methodologies
+getMethodologyById (SqlId methodologyId) = runSelectReturningOne' $
+  lookup_ (esTestMethodology ednaSchema) $ TestMethodologyId $ fromIntegral methodologyId
```
gromakovsky: AFAIU, here `fromIntegral` just applies `SqlSerial` constructor. If that's correct, I think changing `fromIntegral` to `SqlSerial` will make this code more descriptive and shorter at the same time.

backend/src/Edna/Upload/DB/Query.hs
```diff
@@ -0,0 +1,79 @@
...
+
+import Edna.DB.Integration (runInsert', runInsertReturningList', runInsertReturningOne')
+import Edna.DB.Schema
+  (EdnaSchema(..), ExperimentT(..), MeasurementT(..), RemovedMeasurementsT(..), SubExperimentT(..),
+  ednaSchema, theOnlyAnalysisMethodId)
+import Edna.ExperimentReader.Types as EReader
+import Edna.Setup (Edna)
+import Edna.Upload.DB.Schema (ExperimentFileT(..))
+
+insertExperiment :: Word32 -> Word32 -> Word32 -> Edna Word32
```
gromakovsky: Erm, this type looks very said, it has 3 `Word32` arguments that are easy to confuse and returns `Word32` as well. Let's use `CompoundId` and `TargetId` please. In #42 I am adding `ExperimentId`, let's see what will be merged first.

backend/src/Edna/Upload/Service.hs
```diff
@@ -48,20 +45,11 @@ parseFile' = measurementsToSummary . fcMeasurements
 
 compoundNameToId :: Text -> Edna (Maybe (SqlId 'U.CompoundId))
 compoundNameToId compoundName =
-  fmap (fmap mkSqlId) . runSelectReturningOne' $ select $ do
-    compound <- all_ (esCompound ednaSchema)
-    guard_ (cName compound ==. val_ compoundName)
-    return $ cCompoundId compound
+  (SqlId . unSerial . cCompoundId) <<$>> LQ.getCompoundByName compoundName
```
gromakovsky: Since you are doing some refactoring, I propose to define `SqlSerial Word32 -> SqlId anything` helper somewhere.

gromakovsky (changes requested):

backend/src/Edna/Library/DB/Query.hs
gromakovsky: Also, since this function is now exported, it should have a comment about this detail. Here the argument "it's obvious what this function does" doesn't work because inserting a target is not trivial because duplicate names are not allowed. So we should say that in case of a conflict this function does nothing and simply returns the existing ID.

And in general please try to explain what you are doing. Usually I see almost no comments, no PR description, no commit messages.

sashasashasasha151: >Pairing ID with name sounds like something the DB layer doesn't need to care about.

DB layer should construct types which are used by Service layer (Domain values) and Service layer should only work with these Domain types and get/pass them from/to API or DB layer. But as I already said it is not a panacea and for example for Library module we would have to write Domain types and two converters: from/to DB types and from/to API types. Instead of this we just skip them and use only API and DB types.

So, I just want to say that splitting our logic on three layers is already good enough and juggling with types looks a bit over engineering here (I do not suggest to refuse them, but to not think a lot about them). But I agree that returning only IDs is better

backend/src/Edna/Upload/DB/Query.hs
sashasashasasha151: I also added `ExperimentId`, looks like it will be easy to rebase it

backend/src/Edna/Library/DB/Query.hs
```diff
@@ -148,19 +164,23 @@ insertCompound compoundName = do
...
 getMethodologyByName name = runSelectReturningOne' $ select $ do
   methodologies <- all_ $ esTestMethodology ednaSchema
   guard_ (LDB.tmName methodologies ==. val_ name)
   pure methodologies
 
+-- | Get all methodologies
 getMethodologies :: Edna [TestMethodologyRec]
 getMethodologies = runSelectReturningList' $ select $ all_ $ esTestMethodology ednaSchema
 
+-- | Insert methodology and return its DB value
```
gromakovsky, line 183: I would add that it fails in case methodology with this name already exists. Same for `insertProject`.

gromakovsky (approved):

PR #44, opened 2021-03-30
[EDNA-70] Implement Dashboard API

gromakovsky:
## Description

This PR implements all endpoints from `DashboardEndpoints`.

## Related issue(s)

Resolves https://issues.serokell.io/issue/EDNA-70 and https://issues.serokell.io/issue/EDNA-32

- Tests
        silently reappearing again.

- Documentation


backend/sql/drop.sql
```diff
@@ -6,5 +6,6 @@ drop table if exists experiment_file cascade;
 drop table if exists experiment cascade;
 drop table if exists measurement cascade;
 drop table if exists analysis_method cascade;
-drop table if exists subexperiment cascade;
+drop table if exists sub_experiment cascade;
```
sashasashasasha151: Does anyone use this `drop.sql` and what for? (Just a question, I am ok with this file)

backend/src/Edna/Dashboard/DB/Query.hs
```diff
@@ -0,0 +1,154 @@
...
+-- Previous primary sub-experiment is no longer primary.
+makePrimarySubExperiment :: SubExperimentId -> Edna ()
+makePrimarySubExperiment (SqlId subExpId) =
+  runUpdate' $ update (esPrimarySubExperiment ednaSchema)
+  (\pse -> pseSubExperimentId pse <-. val_ subExpId)
+  (\pse -> pseExperimentId pse ==. subquery_ (do
+    subExperiment <- all_ (esSubExperiment ednaSchema)
+    guard_ (seSubExperimentId subExperiment ==. val_ (SqlSerial subExpId))
+    return (seExperimentId subExperiment)
+    ))
```
sashasashasasha151, line 46: 👍

backend/src/Edna/Dashboard/DB/Query.hs
```diff
@@ -0,0 +1,154 @@
...
+  runUpdate' $ update (esSubExperiment ednaSchema)
+  (\se -> seIsSuspicious se <-. val_ isSuspicious)
+  (\se -> seSubExperimentId se ==. val_ (SqlSerial subExpId))
+
+-- | Delete a sub-experiment with given ID along with all @removed_measurement@
+-- entries for this sub-experiment.
+deleteSubExperiment :: SubExperimentId -> Edna ()
+deleteSubExperiment (SqlId subExpId) =
+  runDelete' $ delete (esSubExperiment ednaSchema) $
+  \se -> seSubExperimentId se ==. val_ (SqlSerial subExpId)
```
sashasashasasha151: I thought that we discussed it and if sub experiment is primary we will not delete it 🤔

backend/src/Edna/Dashboard/DB/Query.hs
```diff
@@ -0,0 +1,154 @@
...
+    experiment <- all_ $ esExperiment ednaSchema
+
+    experimentFile <- all_ $ esExperimentFile ednaSchema
+    guard_ (eExperimentFileId experiment ==. cast_ (efExperimentFileId experimentFile) int)
+
+    primarySubExp <- all_ $ esPrimarySubExperiment ednaSchema
+    guard_ (pseExperimentId primarySubExp ==. cast_ (eExperimentId experiment) int)
+
+    subExperiment <- leftJoin_ (all_ $ esSubExperiment ednaSchema) $
+      \subExp -> cast_ (seExperimentId subExp) int ==. eExperimentId experiment
```
sashasashasasha151, line 96: Why not just guard?

backend/src/Edna/Dashboard/DB/Query.hs
```diff
@@ -0,0 +1,154 @@
...
+        )
+      , seSubExperimentId subExperiment
+      )
+  where
+    groupSubExps ::
+      [((ExperimentRec, Word32, Maybe Word32, LocalTime, Word32), Maybe (SqlSerial Word32))] ->
+      [((ExperimentRec, Word32, Maybe Word32, LocalTime, Word32), [SqlSerial Word32])]
+    groupSubExps =
+      map (\ne -> (fst (head ne), mapMaybe snd (toList ne))) .
+      NE.groupBy (\a b -> a ^. _1 . _1 == b ^. _1 . _1)
```
sashasashasasha151: As I remember `groupBy` works only on sorted lists, so maybe add ordered by to query?

backend/src/Edna/Dashboard/DB/Query.hs
```diff
@@ -0,0 +1,154 @@
...
+
+    subExperiment <- leftJoin_ (all_ $ esSubExperiment ednaSchema) $
+      \subExp -> cast_ (seExperimentId subExp) int ==. eExperimentId experiment
+
+    whenJust mProj $ \(SqlId projId) ->
+      guard_ (efProjectId experimentFile ==. val_ projId)
+    whenJust mComp $ \(SqlId compId) ->
+      guard_ (eCompoundId experiment ==. val_ compId)
+    whenJust mTarget $ \(SqlId targetId) ->
+      guard_ (eTargetId experiment ==. val_ targetId)
```
sashasashasasha151: You can leave it this way, but looks like using one filter is easier and more understandable

backend/src/Edna/Web/Server.hs
```diff
@@ -73,6 +74,7 @@ ednaToHandler ctx action =
   `catch` throwPurely @UploadError
   `catch` throwPurely @UploadApiError
   `catch` throwPurely @LibraryError
+  `catch` throwPurely @DashboardError
```
sashasashasasha151, line 77: Why do we need to list all errors here?

backend/src/Edna/Dashboard/Service.hs
```diff
@@ -0,0 +1,130 @@
...
+import Edna.Util
+  (CompoundId, IdType(..), ProjectId, SubExperimentId, TargetId, justOrThrow, unSqlId)
+import Edna.Web.Types (WithId(..))
+
+-- | Make given sub-experiment the primary one for its parent experiment.
+-- Previous primary sub-experiment is no longer primary.
+makePrimarySubExperiment :: SubExperimentId ->
+  Edna (WithId 'SubExperimentId SubExperimentResp)
+makePrimarySubExperiment subExpId = do
+  logMessage $ fmt $ "Making sub-experiment " +| subExpId |+ " primary"
```
sashasashasasha151, line 41: We already log request with responses, looks like no need to duplicate it

backend/src/Edna/Dashboard/Service.hs
```diff
@@ -0,0 +1,130 @@
...
+
+-- | Get data about all experiments using 3 optional filters: by project ID,
+-- compound ID and target ID. If filters by compound and target are specified,
+-- also compute average IC50 for them.
+getExperiments :: Maybe ProjectId -> Maybe CompoundId -> Maybe TargetId ->
+  Edna ExperimentsResp
+getExperiments mProj mComp mTarget = do
+  pairs <- Q.getExperiments mProj mComp mTarget
+  meanIC50 <- case (mComp, mTarget) of
+    (Just _, Just _) -> Just <$> computeMeanIC50 (map snd pairs)
```
sashasashasasha151, line 79: What mean `MeanIC50` for scientists?

sashasashasasha151 (changes requested):

backend/sql/drop.sql
gromakovsky: I often do `\i sql/drop.sql; \i sql/init.sql;` in `psql`, so yes. One day someone will simplify it. Maybe I'll even do it in this PR :)

backend/src/Edna/Dashboard/DB/Query.hs
gromakovsky: Yes, but currently it's checked by a separate SQL query in `Service` module (there we do `isPrimarySubExperiment` check). I decided to postpone it to [EDNA-76](https://issues.serokell.io/issue/edna-76). Though, if I manage to implement it quickly, I'll do it here soon.

backend/src/Edna/Dashboard/DB/Query.hs
gromakovsky: Do you mean inner join instead of left? Currently experiment always has at least one sub-experiment, so we can use inner join. My reasoning was that maybe one day we'll change something and it will be possible for an experiment to have no sub-experiments, in which case left join will be needed here.

backend/sql/drop.sql
sashasashasasha151: Why don't you use setting in config?

gromakovsky: I use setting in config when I need to launch `edna-server`, but sometimes I want to do something with DB directly in which case I do `\i` in `psql`.

backend/src/Edna/Dashboard/DB/Query.hs
gromakovsky: What do you mean? `Maybe (ProjectId, TargetId, CompoundId)`? We need 3 different filters due to our UI/UX.

backend/src/Edna/Web/Server.hs
gromakovsky: Because if you throw an exception, Servant doesn't know how to handle it and treats it as an internal server error, returning the corresponding HTTP code. We want to map all these errors to something different from 500, so we need to catch them and let Servant know what to do with them. And all errors not listed here will be treated as 500.

backend/src/Edna/Dashboard/Service.hs
gromakovsky: 1. We log them only if logging is `Dev`. And if logging mode is `Prod`, my idea was to log some important actions briefly. And here I tentatively treat only mutating actions (i. e. not getters) as _important_. So in `Prod` mode you will see state changes, but won't see all requests, so the log won't be too big.
2. In `Dev` mode (which I treat as `Debug` mode) this duplication may be not bad because sometimes weird things happen, e. g. an endpoint is called, but no state is changed, and these messages help you understand whether you reached `Q.makePrimarySubExperiment`. Also they are easier to read than full requests because they are written in human language.

That said, I see your point too, so if I didn't convince you that these messages are ok, I'll delete them.

backend/src/Edna/Dashboard/Service.hs
gromakovsky: AFAIU the person who decides whether a compound is good or not is interested in IC50 value for `(compound, target that is supposed to be attacked by this compound)`. If they select this compound and target in dashboard, they will see potentially many experiments with this pair and values may be different in these experiments (AFAIU in reality values should be similar, but they can't be absolutely equal). But it's hard to make a decision when you see 10 values, so we also provide average value.

I might be wrong, but I vaguely remember "mean IC50" being mentioned on some meetings, and that's how it appeared in designs.

backend/src/Edna/Dashboard/DB/Query.hs
sashasashasasha151: You can write smth like 

```haskell
filter_ (\(yourLarge tuple) -> eExperimentFileId experiment ==. cast_ (efExperimentFileId experimentFile) int &&.
pseExperimentId primarySubExp ==. cast_ (eExperimentId experiment) int &&
bool (val_ True) (efProjectId experimentFile ==. val_ projId) isJust mProj &&.
bool (val_ True) (eCompoundId experiment ==. val_ compId) isJust mComp &&.
bool (val_ True) (eTargetId experiment ==. val_ targetId) isJust mTarget &&. )
```

But for real I am ok with both solutions, just what you prefer

backend/src/Edna/Web/Server.hs
sashasashasasha151: I think I know the solution how to not write it here but will check later

backend/src/Edna/Dashboard/Service.hs
sashasashasasha151: >So in Prod mode you will see state changes, but won't see all requests, so the log won't be too big.

As I already wrote, on prod it is also very useful to have full logs

>I'll delete them

Lets leave them to merge faster

sashasashasasha151 (approved):

backend/src/Edna/Dashboard/Service.hs
gromakovsky: > As I already wrote, on prod it is also very useful to have full logs

Yes, actually, it's set to Dev in both staging and demo for now: https://github.com/serokell/gemini-infra/pull/10/files
I don't have a real use case for Prod for now, the time will show whether it's useful or not.

PR #46, opened 2021-03-31
Remove capture from upload file

sashasashasasha151:
## Related issue(s)

https://issues.serokell.io/issue/EDNA-60

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/backend/src/Edna/Upload/Web/API.hs
+++ b/backend/src/Edna/Upload/Web/API.hs
@@ -15,19 +15,18 @@ import Universum
 import Data.Aeson.TH (deriveToJSON)
 import Data.Swagger (ToSchema(..))
 import Fmt (Buildable(..))
-import Servant.API (Capture, JSON, Post, QueryParam, Summary, (:>))
+import Servant.API (JSON, Post, Summary, (:>))
 import Servant.API.Generic (AsApi, ToServant, (:-))
-import Servant.Multipart (FileData(..), Mem, MultipartData(..), MultipartForm)
+import Servant.Multipart (Mem, MultipartForm)
 import Servant.Server.Generic (AsServerT, genericServerT)
 import Servant.Util.Combinators.Logging (ForResponseLog(..))
 
 import Edna.ExperimentReader.Parser (parseExperimentXls)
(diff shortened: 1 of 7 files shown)


backend/src/Edna/Upload/Web/API.hs
```diff
@@ -26,8 +29,19 @@ import Edna.ExperimentReader.Types (FileContents(..), Measurement(..), TargetMea
...
 import Edna.Upload.Service (parseFile, uploadFile)
-import Edna.Upload.Web.Types (FileSummary)
-import Edna.Util (MethodologyId, ProjectId, ednaAesonWebOptions, gDeclareNamedSchema)
+import Edna.Upload.Web.Types (FileBS(..), FileSummary, FileUploadReq(..))
+import Edna.Util (SqlId(..), ednaAesonWebOptions, gDeclareNamedSchema)
+
+instance FromMultipart Mem FileBS where
+  fromMultipart = first pretty . expectOneFile
+
+instance FromMultipart Mem FileUploadReq where
```
gromakovsky: Why are these instances here and not in the place where `FileBS` is defined?

backend/src/Edna/Upload/Web/API.hs
```diff
@@ -14,10 +15,12 @@ import Universum
 
 import Data.Aeson.TH (deriveToJSON)
 import Data.Swagger (ToSchema(..))
-import Fmt (Buildable(..))
-import Servant.API (Capture, JSON, Post, QueryParam, Summary, (:>))
+import Data.Text (unpack)
```
gromakovsky: I prefer `toString` because it's easier to understand what is converted to what (`unpack` also exists for bytestrings for example), but that's up to you.

gromakovsky (changes requested):
Looks ok except for one minor thing (placement of instances). Also I hope you've tested that it works when you call it from frontend AND when you call it from swagger UI.

> Problem: there is no way to combine ReqBody with MultipartForm

Btw, it's not the problem we are solving here. We are already not using `ReqBody` with `MultipartForm`. The main problem we are solving, in my opinion, is that description can be big and currently we put it into URL, and very big URLs are not good I think.

gromakovsky (approved):

PR #49, opened 2021-04-05
Compute ic50

sashasashasasha151:
Problem: we need to use python to use it for analysis computing

Solution: for managing python packages and running it use poetry.
Then simply run it from Haskell code using createProcess

## Related issue(s)

https://issues.serokell.io/issue/EDNA-71

- Tests
        silently reappearing again.

- Documentation


backend/app/Main.hs
```diff
@@ -15,6 +18,12 @@ import Edna.Web.Server (edna)
 
 main :: IO ()
 main = withUtf8 $ do
+  (_, Just hout, _, h) <-  createProcess (proc "poetry" ["run", "python", "analysis" </> "ic50.py"])
+    { cwd = Just $ ".." </> "analysis", std_out = CreatePipe}
```
balsoft: This doesn't look right. Running `poetry` will require network access on the host, which is not guaranteed (IIUC the intended use-case is to run Edna in an isolated environment). Also, `..` means the call depends on CWD and the fact that `analysis` dir is present in the parent directory, which will require some flimsy hacks to work in production.

```suggestion
  analysisDir <- fromMaybe (".." </> "analysis") (lookupEnv "ANALYSIS_DIR")
  (_, Just hout, _, h) <-  createProcess (proc "python3" ["analysis" </> "ic50.py"])
    { cwd = Just analysisDir, std_out = CreatePipe}
```

For development, you can just run the Haskell process inside `poetry shell`; In production, we'll run the process inside the environment generated by `poetry2nix`, and with `ANALYSIS_DIR` pointing to the correct place.

balsoft (dismissed):
(the suggestion wasn't tested, it probably needs some changes to actually compile)

analysis/analysis/ic50.py
```diff
@@ -0,0 +1,4 @@
+import numpy as np
```
gromakovsky, line 4: Perhaps it should go to `backend/` because it's part of the backend.

sashasashasasha151: I am not sure that it is good solution. For MVP maybe yes, but in the future as I understand we will have much more python code\
So, I just thought that calling it not from root folder is not a problem

gromakovsky: > For MVP maybe yes, but in the future as I understand we will have much more python code

And what's the problem? I mean `backend/analysis`. So we add just one folder to `backend` which I think is not a big deal, is it?

backend/default.nix
```diff
@@ -13,16 +13,26 @@ let
           ghcOptions = ["-Werror"];
 
           # Do not depend on GHC
```
gromakovsky: Perhaps this comment should be moved one line below.

analysis/analysis/ic50.py
gromakovsky: So what do you think? No matter how much python code we will have, it will still be backend code, no?

backend/edna.cabal
```diff
@@ -82,13 +82,15 @@ library
...
     data-default,
+    filepath,
     fmt,
     microlens,
     microlens-platform,
     network-uri,
     network-uri-json,
     optparse-applicative,
     postgresql-simple,
+    process,
```
gromakovsky, line 93: I saw in Slack that cool boyz use `typed-process`, did you consider using it?

backend/src/Edna/Analysis/FourPL.hs
```diff
@@ -45,8 +60,49 @@ instance ToSchema Params4PL where
...
+
+data Params4PLResp = Params4PLResp
+  { plrspExperiment :: ExperimentId
+  , plrspData :: AnalysisResult
+  } deriving stock (Generic, Show, Eq)
+
+deriveJSON ednaAesonWebOptions ''Params4PLResp
+
+-- | This function performs actual analysis and finds the best parameters.
+analyse4PL :: [(ExperimentId, [Measurement])] -> Edna [Params4PLResp]
```
gromakovsky: 1. Why does this module need to know about `ExperimentId`s at all? Why can't it take a list of lists of measurements (maybe wrapped into a newtype) and return a list of `AnalysisResult`s?
2. Why `ExperimentId` rather than `SubExperimentId`?

backend/src/Edna/Analysis/FourPL.hs
```diff
@@ -45,8 +60,49 @@ instance ToSchema Params4PL where
...
+analyse4PL experiments = do
+  let request = unpack $ encode $ flip map experiments $
+        \(eId, ms) -> Params4PLReq eId $ map (\m -> (mConcentration m, mSignal m)) ms
+  logMessage $ T.pack $ "Python data request: " <> request
+  analysisDir <- liftIO $ fromMaybe (".." </> "analysis") <$> (lookupEnv "ANALYSIS_DIR")
```
gromakovsky: I would put details of calling `python` into a separate helper function (which later may move into a separate module).

backend/src/Edna/Analysis/FourPL.hs
```diff
@@ -45,8 +60,49 @@ instance ToSchema Params4PL where
...
+  } deriving stock (Generic, Show, Eq)
+
+deriveJSON ednaAesonWebOptions ''Params4PLResp
+
+-- | This function performs actual analysis and finds the best parameters.
+analyse4PL :: [(ExperimentId, [Measurement])] -> Edna [Params4PLResp]
+analyse4PL experiments = do
+  let request = unpack $ encode $ flip map experiments $
+        \(eId, ms) -> Params4PLReq eId $ map (\m -> (mConcentration m, mSignal m)) ms
+  logMessage $ T.pack $ "Python data request: " <> request
```
gromakovsky: 1. AFAIU the request may be quite big, so I think `logDebug` is more appropriate (I think you rarely need to see full requests, only when something goes wrong in analysis).
2. I prefer `toText`/`toString` functions, but that's up to you. I just don't remember in which direction `pack` converts and in which `unpack`.

backend/src/Edna/Analysis/FourPL.hs
```diff
@@ -45,8 +60,49 @@ instance ToSchema Params4PL where
...
+  (_, Just hout, Just herr, h) <- liftIO $ createProcess
+    (proc "python3" ["analysis" </> "ic50.py", request])
+    { cwd = Just analysisDir, std_out = CreatePipe, std_err = CreatePipe}
+  _ <- liftIO $ waitForProcess h
+
+  err <- liftIO $ hGetContents herr
+  unless (null err) $ logMessage $ "Unexpected python errors: " <> err
+
+  out <- liftIO $ hGetContents hout
+  let analysisResults = eitherDecode $ pack $ toString out
```
gromakovsky: `Data.ByteString.Lazy.Char8` functions look dangerous to me and I think it shouldn't be used here. If there is a non-ASCII character in stdout for some reason (it seems unlikely, but who knows), you will get something weird. I think you want `encodeUtf8` here.

backend/src/Edna/Analysis/FourPL.hs
```diff
@@ -45,8 +60,49 @@ instance ToSchema Params4PL where
...
+deriveJSON ednaAesonWebOptions ''Params4PLResp
+
+-- | This function performs actual analysis and finds the best parameters.
+analyse4PL :: [(ExperimentId, [Measurement])] -> Edna [Params4PLResp]
+analyse4PL experiments = do
+  let request = unpack $ encode $ flip map experiments $
+        \(eId, ms) -> Params4PLReq eId $ map (\m -> (mConcentration m, mSignal m)) ms
+  logMessage $ T.pack $ "Python data request: " <> request
+  analysisDir <- liftIO $ fromMaybe (".." </> "analysis") <$> (lookupEnv "ANALYSIS_DIR")
+  (_, Just hout, Just herr, h) <- liftIO $ createProcess
```
gromakovsky: Why not `readCreateProcessWithExitCode`?

backend/src/Edna/Dashboard/Service.hs
```diff
@@ -87,19 +87,20 @@ getExperiments mProj mComp mTarget = do
 -- but let's not optimize prematurely.
 computeMeanIC50 :: [ExperimentResp] -> Edna Double
 computeMeanIC50 resps = do
-  avg <$> mapM (getDefaultIC50 . erPrimarySubExperiment) resps
+  avg . foldr (\e es -> case e of {Left _ -> es; Right x -> (p4plC x) : es}) [] <$>
```
gromakovsky: Looks like `mapMaybe` can be used to simplify it.

backend/src/Edna/Orphans.hs
```diff
@@ -15,6 +15,12 @@ import Servant.Util.Common.Common (ApiHasArgClass(..))
 -- so it's simpler to just have one line for now.
 deriving newtype instance MonadCatch (RIO env)
 
+-- It's better to use @throw@ family of functions instead of @fail@, but
+-- in tests we want to do incomplete pattern-matching sometimes and @MonadFail@
+-- is needed for that.
+instance MonadFail (RIO env) where
```
gromakovsky: Looks like, if you use `readCreateProcessWithExitCode`, this instance won't be needed outside of tests.

gromakovsky (changes requested):

backend/edna.cabal
sashasashasasha151: Yes, but only if we will have time. So, if I will rebase and implement all other fixes before demo and we will have time to use `typed-process` I will add it

backend/src/Edna/Analysis/FourPL.hs
sashasashasasha151: 1. Hm, for real it doesn't need it. I just wanted to add some info to responses/requests and also to guarantee that it will be in the origin sequence (like when we have ids we can not think about it)
2. Because we don't know `ExperimentId` before computing it

So, I don't see any disadvantages to not use `ExperimentId` here

analysis/analysis/ic50.py
sashasashasasha151: 1. I don't want to see python code in stack project. For example in the future it will be big enough and it will be strange as for me
2. There are no advantages putting it in backend folder and also no disadvantages 
3. If edna will be big enough I think python part will be fully separate from haskell part. So, maybe I am wrong but I think it will work as a separate server. So, haskell will just call some api to `analysis` server and get result. In. this case, combining this things will be a little bit :ghc:

backend/src/Edna/Analysis/FourPL.hs
sashasashasasha151: Lets leave to times when we will use or not `typed-process` (but if it is easy to fix I will do it)

backend/default.nix
sashasashasasha151: So, I moved it, but not really understand why

gromakovsky: Because it belongs to `dontStrip = false;`, doesn't it?

backend/src/Edna/Analysis/FourPL.hs
gromakovsky: It was trivial to refactor, so I went ahead and did it myself.

gromakovsky: Also added a check for `ExitCode` because I think it matters (if non-zero code is returned, something is wrong and we should report it).

backend/src/Edna/Analysis/FourPL.hs
```diff
@@ -49,8 +63,49 @@ instance ToSchema Params4PL where
...
+  { plrspExperiment :: ExperimentId
+  , plrspData :: AnalysisResult
+  } deriving stock (Generic, Show, Eq)
+
+deriveJSON ednaAesonWebOptions ''Params4PLResp
+
+callPythonAnalysis :: String -> Edna (Either String [Params4PLResp])
+callPythonAnalysis request = do
+  logDebug $ toText $ "Python data request: " <> request
+  analysisDir <- liftIO $ fromMaybe (".." </> "analysis") <$> (lookupEnv "ANALYSIS_DIR")
```
gromakovsky:
```suggestion
  analysisDir <- liftIO $ fromMaybe (".." </> "analysis") <$> lookupEnv "EDNA_ANALYSIS_DIR"
```

Please rename `ANALYSIS_DIR` to `EDNA_ANALYSIS_DIR` here and in nix files to avoid possible name collisions. Also brackets look redundant.

analysis/analysis/ic50.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def main(argv):
+    data = json.loads(argv[0])
+
+    response = []
+
+    for value in data:
+        result = compute_sigmoid(np.array(value["data"], dtype=np.float32))
+        response_value = {"experiment": value["experiment"], "data": {"Right": result}}
```
gromakovsky, line 59: I didn't review this file carefully because I hope @HellyHenry will do it, but this `Right` looks suspicious? Do you always return success? Can't the above process fail for some reason? If it can, what happens when it fails? If it can't, isn't this `Right` redundant? Is it reserved for possibility of errors in future?

backend/src/Edna/Analysis/FourPL.hs
```diff
@@ -49,8 +63,49 @@ instance ToSchema Params4PL where
   declareNamedSchema Proxy =
     declareNamedSchema @Params4PLTuple Proxy
 
+type AnalysisResult = Either Text Params4PL
+
+instance Buildable (AnalysisResult) where
+  build = build . (coerce :: AnalysisResult -> Either Text Params4PL)
```
gromakovsky: Why do we need this `coerce`? Looks like `. id`.

backend/src/Edna/Dashboard/Service.hs
```diff
@@ -144,19 +145,20 @@ getExperiments mProj mComp mTarget = do
...
 computeMeanIC50 resps = do
-  avg <$> mapM (getDefaultIC50 . erPrimarySubExperiment) resps
+  avg . mapMaybe (rightToMaybe . second p4plC) <$>
+    mapM (getDefaultIC50 . erPrimarySubExperiment) resps
   where
     avg :: [Double] -> Double
     avg items = sum items / fromIntegral (length items)
 
-    getDefaultIC50 :: SubExperimentId -> Edna Double
+    getDefaultIC50 :: SubExperimentId -> Edna AnalysisResult
```
gromakovsky: It should be renamed to `getDefaultResult` then or something like that.

backend/src/Edna/Upload/DB/Query.hs
```diff
@@ -27,20 +28,37 @@ import Edna.Util
...
 
 -- | Insert experiment and return its ID
-insertExperiment :: ExperimentFileId -> CompoundId -> TargetId -> Edna ExperimentId
-insertExperiment (SqlId experimentFileId) (SqlId compoundId) (SqlId targetId) =
-  fromSqlSerial . eExperimentId <$>
-    runInsertReturningOne' (insert (esExperiment ednaSchema) $ insertExpressions
-      [ ExperimentRec
+insertExperiments :: [(ExperimentFileId, CompoundId, TargetId)] -> Edna [ExperimentId]
+insertExperiments experiments = map (fromSqlSerial . eExperimentId) .
+  sortBy (\ a b -> (eCompoundId a, eTargetId a) `compare` (eCompoundId b, eTargetId b)) <$>
```
gromakovsky: Why not `sortOn` or `sortWith`?

backend/src/Edna/Upload/DB/Query.hs
```diff
@@ -27,20 +28,37 @@ import Edna.Util
...
         , eExperimentFileId = val_ experimentFileId
         , eCompoundId = val_ compoundId
         , eTargetId = val_ targetId
-        }
-      ])
+        })
+
+-- | Insert primary sub-experiment and return its ID
+--
+-- TODO [EDNA-73] Support non-primary sub-experiments
```
gromakovsky: Looks like bad rebase, this comment should be gone.

backend/test/Test/DB/Gen.hs
```diff
@@ -93,7 +93,7 @@ genSubExperimentRec :: Gen.MonadGen m => Word32 -> Word32 -> Word32 -> m SubExpe
 genSubExperimentRec subExperimentId seAnalysisMethodId seExperimentId = do
   seName <- genName
   seIsSuspicious <- Gen.bool
-  seResult <- PgJSON <$> genParams4PL
+  seResult <- PgJSON <$> Gen.either genName genParams4PL
```
gromakovsky, line 96: Using `genName` for errors is :joy: but ok.

gromakovsky:
Some additional items (maybe they can go into a separate PR):
1. Update usage instructions to say how to deal with `poetry`. Partially overlaps with #40.
2. Update `Makefile` to provide Poetry environment to `make test`.
3. Maybe we need a script to run `edna-server` inside Poetry environment because now it's non-trivial.
4. Check whether Poetry environment is available on start. AFAIU if we run `edna-server` incorrectly (with regards to Python/Poetry), it will start successfully, but will not be able to do any actions that involve analysis.

backend/src/Edna/Dashboard/Service.hs
```diff
@@ -98,20 +98,21 @@ newSubExperiment subExpId req = do
...
+      [WithId 'MeasurementId MeasurementResp] -> [Measurement]
     computeNewPoints = mapMaybe stepActive
 
     stepActive ::
-      WithId 'MeasurementId MeasurementResp -> Maybe (Double, Double)
-    stepActive wi@WithId {..} =
-      (mrConcentration wItem, mrSignal wItem) <$ guard (not $ isRemoved wi)
+      WithId 'MeasurementId MeasurementResp -> Maybe Measurement
+    stepActive wi@WithId {wItem = MeasurementResp{..}} =
+      Measurement mrConcentration mrSignal mrIsEnabled <$ guard (not $ isRemoved wi)
```
gromakovsky: So you set `mIsOutlier` to `mrIsEnabled` here, it looks strange to me. This value is eventually ignored, so you can set it to arbitrary value, but overall it looks like a hack to me. If I read it correctly, you are using a type that stores more data than necessary and set unused data to arbitrary values. I think we should distinguish measurement that it just 2 values and measurement which is 2 values + flag whether it's outlier.

analysis/analysis/ic50.py
sashasashasasha151: Its limitation that we have now as we don't handle fitting errors. I suggest to do it in a separate pr if we want to show demo. So, for now `Left` is impossible and all errors we put to console

backend/src/Edna/Analysis/FourPL.hs
sashasashasasha151: I don't understand why I need this instance \
Looks like it do nothing as its `type`, not `newtype` \
But without this instance it doesn't compile

backend/src/Edna/Analysis/FourPL.hs
gromakovsky: > So, I don't see any disadvantages to not use `ExperimentId` here

1. You need to call `getExperimentId` in `analyseNewSubExperiment`. It's not expensive, but still redundant.
2. I still find it confusing that `ExperimentId`s are used here because we do analysis for sub-experiments, not for experiments.

That said, probably it can be put into "to be handled later" group, maybe I'll refactor it myself in a separate PR.

backend/src/Edna/Analysis/FourPL.hs
```diff
@@ -66,7 +65,7 @@ instance ToSchema Params4PL where
 type AnalysisResult = Either Text Params4PL
 
 instance Buildable (AnalysisResult) where
-  build = build . (coerce :: AnalysisResult -> Either Text Params4PL)
+  build = build
```
gromakovsky: And now it's infinite recursion.

backend/src/Edna/Upload/DB/Query.hs
```diff
@@ -26,21 +27,36 @@ import Edna.Util
...
         , eCompoundId = val_ compoundId
         , eTargetId = val_ targetId
-        }
-      ])
+        })
+
+-- | Insert sub-experiments and return its IDs
+insertSubExperiments :: [Params4PLResp] -> Edna [SubExperimentId]
+insertSubExperiments params = map (fromSqlSerial . seSubExperimentId) .
+  sortBy (\ a b -> seExperimentId a `compare` seExperimentId b) <$>
```
gromakovsky: There are more places (this one + 2) where `sortBy` is used with `compare` inside and apparently can be simplified using `sortWith` or `sortOn`.

backend/src/Edna/Analysis/FourPL.hs
```diff
@@ -49,8 +62,47 @@ instance ToSchema Params4PL where
   declareNamedSchema Proxy =
     declareNamedSchema @Params4PLTuple Proxy
 
+type AnalysisResult = Either Text Params4PL
+
+instance Buildable (AnalysisResult) where
```
gromakovsky:
```suggestion
instance Buildable AnalysisResult where
```

backend/src/Edna/Analysis/FourPL.hs
gromakovsky: > I don't understand why I need this instance 

What do you mean? You need this instance because you have
```
instance Buildable (ForResponseLog AnalysisResult) where
  build = buildForResponse
```

and this `buildForResponse` requires `instance Buildable AnalysisResult` which you need to define.

analysis/analysis/ic50.py
gromakovsky: > I don't want to see python code in stack project. For example in the future it will be big enough and it will be strange as for me

But then we can have `backend/server` and `backend/analysis` for example. Or `backend/haskell` and `backend/python`.

Anyway, I don't want to spend more time on this discussion, let's leave it as is and come back to it later when we have more python code.

backend/src/Edna/Upload/DB/Query.hs
```diff
@@ -26,21 +27,36 @@ import Edna.Util
   (CompoundId, ExperimentFileId, ExperimentId, MeasurementId, MethodologyId, ProjectId, SqlId(..),
   SubExperimentId, TargetId, fromSqlSerial)
 
--- | Insert experiment and return its ID
-insertExperiment :: ExperimentFileId -> CompoundId -> TargetId -> Edna ExperimentId
-insertExperiment (SqlId experimentFileId) (SqlId compoundId) (SqlId targetId) =
-  fromSqlSerial . eExperimentId <$>
-    runInsertReturningOne' (insert (esExperiment ednaSchema) $ insertExpressions
-      [ ExperimentRec
+-- | Insert experiments and return its IDs
```
gromakovsky:
```suggestion
-- | Insert experiments and return their IDs
```

backend/src/Edna/Upload/DB/Query.hs
```diff
@@ -26,21 +27,36 @@ import Edna.Util
...
+        ExperimentRec
         { eExperimentId = default_
         , eExperimentFileId = val_ experimentFileId
         , eCompoundId = val_ compoundId
         , eTargetId = val_ targetId
-        }
-      ])
+        })
+
+-- | Insert sub-experiments and return its IDs
```
gromakovsky:
```suggestion
-- | Insert sub-experiments and return their IDs
```

backend/src/Edna/Upload/Service.hs
```diff
@@ -103,18 +102,21 @@ uploadFile' projSqlId@(SqlId proj) methodSqlId@(SqlId method)
...
-              getId :: HasCallStack => Text -> HashMap Text (SqlId a) -> SqlId a
-              getId name =
-                fromMaybe (error $ "unexpected name: " <> name) .
-                view (at name)
-            let targetId = getId targetName targetToId
-            let compoundId = getId compoundName compoundToId
-            insertExperiment expFileId compoundId targetId measurements
-
+      experiments <- sortBy
+        (\a b -> (a ^. _1 . _2, a ^. _1 . _1) `compare` (b ^. _1 . _2, b ^. _1 . _1)) <$>
```
gromakovsky: I see a lot of sorting in various places, can you explain why we need it?

worm2fed:
I think we can use python typing instruments to make code a bit safer

https://docs.python.org/3/library/typing.html
https://github.com/python/typing

backend/test/Test/DashboardSpec.hs
```diff
@@ -80,34 +80,36 @@ spec = withContext $ do
...
-          step :: Word32 -> Measurement -> WithId 'MeasurementId MeasurementResp
-          step i m = WithId (SqlId i) MeasurementResp
-            { mrConcentration = mConcentration m
-            , mrSignal = mSignal m
-            , mrIsEnabled = (i == changedMeasurement) == mIsOutlier m
+          step :: Word32 -> Measurement -> WithId 'MeasurementId Measurement
+          step i m = WithId (SqlId i) Measurement
+            { mConcentration = mConcentration m
+            , mSignal = mSignal m
+            , mIsOutlier = (i == changedMeasurement) == mIsOutlier m
```
gromakovsky: 1. Why did you change `MeasurementResp` to `Measurement`? Is it still needed?
2. Changing `mIsOutlier` value looks like a hack here in any case. `mIsOutlier` tells us whether this measurement was marked as outlier in source data. Once we read it, we never change it. Unlike `mrIsEnabled` which may differ in different sub-experiments.

backend/test/Test/DashboardSpec.hs
```diff
@@ -158,7 +160,8 @@ gettersSpec = do
           getExperiments (Just $ SqlId 1) (Just compoundId) (Just targetId)
         liftIO $ do
           length erExperiments `shouldBe` 1
-          erMeanIC50 `shouldBe` Just 33
+          -- TODO impossible to compute mean based on python results
+--          erMeanIC50 `shouldBe` Just 33
```
gromakovsky: Here you can run `analyse4PL` for one specific experiment (because there is only one in result) and check that `erMeanIC50` is equal to IC50 computed for that experiment. Can't you?

backend/test/Test/SampleData.hs
```diff
@@ -121,42 +121,48 @@ compoundName5 = "comp5"
...
+m5Outlier = Measurement
   { mConcentration = 6
   , mSignal = 5000
   , mIsOutlier = True
   }
+m6 = m1 {mConcentration = 60, mSignal = 400}
+m7 = m1 {mConcentration = 70, mSignal = 250}
+m8 = m1 {mConcentration = 80, mSignal = 150}
+m9 = m1 {mConcentration = 90, mSignal = 120}
+m10 = m1 {mConcentration = 100, mSignal = 100}
```
gromakovsky, line 165: I see that you added a number of new measurements, but I don't see new tests or substantial changes in existing tests. So what was your goal when you added them?

analysis/pyproject.toml
```diff
@@ -0,0 +1,19 @@
...
+name = "edna"
+version = "0.1.0"
+description = ""
+authors = ["Serokell <hi@serokell.io>"]
+
+[tool.poetry.dependencies]
+python = "3.9.*"
+numpy = "^1.20.2"
+scipy = "^1.6.2"
+matplotlib = "^3.4.1"
```
HellyHenry: Not sure why matplotlib is needed in production

gromakovsky (changes requested):
Okay, I think the only 100% necessary things to merge this PR are:
1. Remove infinite recursion in `Buildable`.
2. Get approval from @HellyHenry regarding `ic50.py`

gromakovsky:
AFAIU there are no tests for fitting on 4 example files, right?

analysis/analysis/ic50.py
```diff
@@ -0,0 +1,65 @@
...
+from sklearn.preprocessing import MinMaxScaler
+
+
+def find_nearest_idx(array, value):
+    array = np.asarray(array)
+    idx = (np.abs(array - value)).argmin()
+    return idx
+
+
+def sigmoid(x, a, b, c, d):
```
HellyHenry, line 15: You can add typing for more safety, it'll take little of your time. 
Here's a good guide on how to do it: https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html
Also you can use MyPy to test for missing static typing

analysis/analysis/ic50.py
```diff
@@ -0,0 +1,65 @@
...
+    return [transformed[0], res1[1], res1[2], transformed[1]]
+
+
+def main(argv):
+    data = json.loads(argv[0])
+
+    response = []
+
+    for value in data:
+        result = compute_sigmoid(np.array(value["data"], dtype=np.float32))
```
HellyHenry, line 58: Not sure, how strict the back-end is in terms of missing response from analyser. I think that perfecting the script requires a lot of testing on different data points and not needed for MVP (before we get some feedback). I'd suggest you can add something like `try` and `except` to capture potential errors and return error description in those cases, just so that the code don't fall. It's a matter of what's expected to be an output. And later proper unit and logic tests will be added to cover the whole data analysis pipeline.

analysis/analysis/ic50.py
```diff
@@ -0,0 +1,65 @@
...
+        [s[1] for s in sorted_measurements])
+
+
+def compute_sigmoid(measurements):
+    concentrations, signals = replace_duplicates(measurements)
+
+    scaler = MinMaxScaler()
+    signals_scalered = scaler.fit_transform(signals.reshape(-1, 1))
+    signals_scalered = signals_scalered.reshape(-1)
+
```
HellyHenry, line 40: You can silence warnings, so that the logs will be empty. Something like this:
`import warnings`
`warnings.filterwarnings("ignore")`

analysis/analysis/ic50.py
```diff
@@ -0,0 +1,65 @@
+import json
```
HellyHenry, line 1: You can run `isort ic50.py` in terminal so that the imports are structured in a nice way.

analysis/analysis/ic50.py
HellyHenry: I suppose you ran tests on different files provided to us, right?

backend/src/Edna/Analysis/FourPL.hs
sashasashasasha151: But `AnalysisResult` is just type synonym

backend/src/Edna/Analysis/FourPL.hs
sashasashasasha151: But it works 🤔

backend/src/Edna/Upload/Service.hs
sashasashasasha151: Because `returning` functionality from sql doesn't guarantee order in which you insert/update values and we need it. There is `with` functionality with which we can sort our `returning`, but beam doesn't support it (it support only `with select`, not `with insert`)

backend/test/Test/DashboardSpec.hs
sashasashasasha151: It was needed when 4pl computing took `Measurement` as param

backend/test/Test/SampleData.hs
sashasashasasha151: Because fitting doesn't work on number of points less then number of params

analysis/analysis/ic50.py
sashasashasasha151: All 4 files work 

> add something like try and except to capture potential errors

So, we want to present demo on this week and I just wanted to create MVP of python code for it and add all this addition improvements to future task. So, if we will have time before demo to do it, I will do it as the most priority task, but if not, will do after demo

analysis/analysis/ic50.py
sashasashasasha151: Hm, as I saw warnings are in stderr, and I think we should show all warnings and errors to catch unexpected behaviour

backend/src/Edna/Analysis/FourPL.hs
gromakovsky: No, did you try calling `/new/analyse` endpoint? It hangs.

backend/test/Test/DashboardSpec.hs
```diff
@@ -160,8 +157,7 @@ gettersSpec = do
           getExperiments (Just $ SqlId 1) (Just compoundId) (Just targetId)
         liftIO $ do
           length erExperiments `shouldBe` 1
-          -- TODO impossible to compute mean based on python results
---          erMeanIC50 `shouldBe` Just 33
+          erMeanIC50 `shouldBe` Just 51.025965961684655
```
gromakovsky, line 160: I think it's more :proper: to actually call analysis here, but that's ok for now, my 33 was not proper either.

backend/src/Edna/Analysis/FourPL.hs
sashasashasasha151: Okay, I tested just `new`, not `/new/analyse`. `build . id` also doesn't work. I maybe don't understand smth, but why I need instance for type synonym?

backend/src/Edna/Analysis/FourPL.hs
gromakovsky: > But `AnalysisResult` is just type synonym

So what? There no `instance Buildable (Either a b)` if that's what you mean.

backend/src/Edna/Analysis/FourPL.hs
gromakovsky: You need instance for `Either Text Params4PL`, it simply doesn't exist.

backend/src/Edna/Upload/Service.hs
gromakovsky: Ok, but my question is: is this sorting just for neater presentation on UI or we rely on it somewhere in backend logic?

sashasashasasha151: Backend

backend/src/Edna/Analysis/FourPL.hs
sashasashasasha151: :fffuuu: \
Fixed

backend/src/Edna/Analysis/FourPL.hs
sashasashasasha151: I thought that it is and didn't understand what is doing on

backend/src/Edna/Upload/Service.hs
gromakovsky: Ok, where?

gromakovsky: I can probably figure it out myself, but it'll be simpler if you explain.

gromakovsky (approved):

backend/src/Edna/Upload/Service.hs
```diff
@@ -103,18 +102,20 @@ uploadFile' projSqlId@(SqlId proj) methodSqlId@(SqlId method)
...
-            let
-              getId :: HasCallStack => Text -> HashMap Text (SqlId a) -> SqlId a
-              getId name =
-                fromMaybe (error $ "unexpected name: " <> name) .
-                view (at name)
-            let targetId = getId targetName targetToId
-            let compoundId = getId compoundName compoundToId
-            insertExperiment expFileId compoundId targetId measurements
-
+      experiments <- sortWith (\a -> (a ^. _1 . _2, a ^. _1 . _1)) <$>
```
gromakovsky: AFAIU `_2, _1` means `compoundId, expFileId`. Don't we want `compoundId, targetId`?

backend/src/Edna/Upload/DB/Query.hs
```diff
@@ -26,21 +27,36 @@ import Edna.Util
...
+        ExperimentRec
         { eExperimentId = default_
         , eExperimentFileId = val_ experimentFileId
         , eCompoundId = val_ compoundId
         , eTargetId = val_ targetId
-        }
-      ])
+        })
+
+-- | Insert sub-experiments and return their IDs
```
gromakovsky, line 44: Would be great to mention sorting in comments.

gromakovsky:
> AFAIU there are no tests for fitting on 4 example files, right?

@sashasashasasha151 what about this?

gromakovsky:
Created a couple of follow-up issues:
1. https://issues.serokell.io/issue/EDNA-85
2. https://issues.serokell.io/issue/EDNA-86

sashasashasasha151:
>what about this?

@gromakovsky what tests do you want? I don't think that it will be informative. Just some hardcoded values which we will change every time we change fitting function or what points we use. Because of this I also just commented `meanIC50` test. Because we got value from python run and just put it to tests. I think we should ask @HellyHenry how real word test such things. So we have some data and have some result based on this data. Now, how to know that this result is good or not 🤔.

It is easy to do when we have some traning + test data of for example when we have some upper bound of error, but when we have nothing it is not understandable for me

gromakovsky:
> @gromakovsky what tests do you want?

I want to check that our results are not worse than sample results that we have. We can compute inaccuracy of sample results and of our results and compare them. If our results are not worse, they are somewhat sane. Maybe they are not ideal, but definitely not total crap.

>  Just some hardcoded values which we will change every time we change fitting function or what points we use.

No, they won't change. We have sample values computed by another tool, they are constant and we know that they are _not too bad_. If we changing something in implementation of fitting, parameters that we find will change, but the test will keep passing as long as we compute more or less correct values.

> Because of this I also just commented meanIC50 test. Because we got value from python run and just put it to tests.

The goal there is to check that the mean value is computed correctly, that it's indeed the mean value of IC50 results from primary sub-experiments. Since there is only one experiment in the result, that test is not very informative, it tests that the mean value of 1 value is the value itself. But even this case I think is useful. Ideally there should be cases with more than 1 experiment having more than one sub-experiment. Maybe I'll add it as part of the new EDNA-85.

Anyway, if there are no tests, it's not a problem as long as you manually tested all 4 files. I'll add tests myself in EDNA-85.

PR #50, opened 2021-04-05
Use HTTPS instead of git URI in stack.yaml

mkaito:
Nix does not accept bare `git@` URIs due to regression: https://github.com/NixOS/nix/issues/3503

### DIFF
--- a/backend/stack.yaml
+++ b/backend/stack.yaml
@@ -32,7 +32,7 @@ extra-deps:
 # with more precise goals.
 # It's developed by us, so we are quite confident in it, that's why we are using
 # it even though it's not ready yet.
-- git: git@github.com:serokell/servant-util.git
+- git: https://github.com/serokell/servant-util
   commit: 3dff2de34cce90c4d1dab68e03d30f8939fd52ca # not master
   subdirs:
     - servant-util
--- a/backend/cabal.project
+++ b/backend/cabal.project
@@ -7,7 +7,7 @@ packages:
 
 source-repository-package
     type: git
-    location: git@github.com:serokell/servant-util.git
+    location: https://github.com/serokell/servant-util
     tag: 3dff2de34cce90c4d1dab68e03d30f8939fd52ca
     subdir: servant-util
 


gromakovsky:
What exactly doesn't work without this change?

gromakovsky:
I mean that `haskell.nix` seems to handle it well :shrug:

backend/stack.yaml
```diff
@@ -32,7 +32,7 @@ extra-deps:
 # with more precise goals.
 # It's developed by us, so we are quite confident in it, that's why we are using
 # it even though it's not ready yet.
-- git: git@github.com:serokell/servant-util.git
+- git: ssh://git@github.com/serokell/servant-util.git
```
balsoft:
```suggestion
- git: https://github.com/serokell/servant-util.git
```

balsoft:
@gromakovsky Before this change, it didn't build on Nix > 2.3.3.

gromakovsky:
Okay, and CI runner in our repos uses ≤ 2.3.3, right? Because `git@github.com:serokell/servant-util.git` format seems to work fine in all repos, most of our Haskell repos have such dependencies.

balsoft:
That's correct, we still have stable Nix on Albali.

gromakovsky (approved):
I approve, but @balsoft suggests using `https` and I don't know what's better.

mkaito:
It doesn't matter. The problem is that nix >2.3 does not recognize `git@github.com` as a valid URL and we want to use something else.

gromakovsky:
Ok, @mkaito @balsoft please reach consensus what URI is better :wink: 

And please update the commit message and explain what happens here.

> The problem is that nix >2.3 does not recognize `git@github.com` as a valid URL and we want to use something else.

It should be written in the commit message :wink:

mkaito:
Fixed and reworded the commit.

As for why we even want to fix this, given that it does not seem to affect stable Nix, the issue is over a year old and we have no guarantees that it will be fixed before 2.4 becomes stable. So we want to future-proof nix evaluation of the project. Just in case.

balsoft (approved):

gromakovsky (approved):

PR #51, opened 2021-04-05
[EDNA-60] Add/extend tests for Library and Upload services

gromakovsky:
## Description

There are no tests for Library and only naive tests for Upload. This PR fixes it by adding more tests.

## Related issue(s)

Finally resolves https://issues.serokell.io/issue/EDNA-60

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/backend/src/Edna/Upload/Web/Types.hs
+++ b/backend/src/Edna/Upload/Web/Types.hs
@@ -4,6 +4,8 @@ module Edna.Upload.Web.Types
   , FileSummaryItem (..)
   , FileBS (..)
   , FileUploadReq (..)
+
+  , sortFileSummary
   ) where
 
 import Universum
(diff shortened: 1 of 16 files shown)


backend/src/Edna/Upload/Web/Types.hs
```diff
@@ -132,6 +134,13 @@ newtype FileSummary = FileSummary
   } deriving stock (Generic, Show, Eq)
     deriving newtype (Buildable)
 
+-- | Deeply sort all items (targets and compounds for each target) in 'FileSummary'.
+sortFileSummary :: FileSummary -> FileSummary
+sortFileSummary (FileSummary items) = FileSummary $ sortOn fsiTarget items'
```
sashasashasasha151, line 139: Do you need it only for tests?

backend/test/Test/Setup.hs
```diff
@@ -45,7 +48,8 @@ withContext = around withContext'
       connString <- postgresTestServerConnString
       let testConfig = defaultEdnaConfig &
             ecDb . dbInitialisation ?~ DbInit EnableWithDrop "./sql/init.sql" &
-            ecDb . dbConnString .~ connString
+            ecDb . dbConnString .~ connString &
+            ecLogging .~ LogNothing
```
sashasashasasha151, line 52: Oh, it #49 it was a pain to see all logs

sashasashasasha151 (approved):

backend/src/Edna/Upload/Web/Types.hs
gromakovsky: No, I also use it in `measurementsToSummary` to return sorted summary from `uploadFile` and `parseFile`. Perhaps it's better for users when summary is sorted, it's easier to find something there.

PR #53, opened 2021-04-07
[EDNA-73] Extend Dashboard API

gromakovsky:
## Description

Problem: currently `edna-server` doesn't expose any way to create a new sub-experiment via its API.
Also there is no way to get experimental data file (it should be possible to download them) and metadata.

Solution:
1. Add a new endpoint to create sub-experiments.
2. Also add a similar one to compute 4PL parameters without saving a new sub-experiment.
3. Update `getMeasurements` to return IDs along with measurements
because sub-experiment creation endpoint takes measurement IDs as
argument.
4. Add endpoints to get metadata and raw experiment data file.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-73

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/backend/src/Edna/Analysis/FourPL.hs
+++ b/backend/src/Edna/Analysis/FourPL.hs
@@ -53,4 +53,4 @@ instance ToSchema Params4PL where
 --
 -- TODO [EDNA-71] Implement!
 analyse4PL :: [(Double, Double)] -> IO Params4PL
-analyse4PL _ = pure (Params4PL 1 2 3 4)
+analyse4PL _ = pure (Params4PL 36404 1.14 33 -2552)
--- a/backend/src/Edna/Analysis/FourPL.hs
+++ b/backend/src/Edna/Analysis/FourPL.hs
@@ -10,6 +10,7 @@ import Universum
 import Data.Aeson (FromJSON(..), ToJSON(..))
 import Data.Swagger (ToSchema(..))
 import Fmt (Buildable(..), tupleF)
+import Servant.Util.Combinators.Logging (ForResponseLog, buildForResponse)
 
 -- | Parameters of 4PL function, analysis outcome of this analysis method.
 -- The function is defined as follows:
@@ -30,6 +31,9 @@ toTuple Params4PL {..} = (p4plA, p4plB, p4plC, p4plD)
 
 instance Buildable Params4PL where
   build = tupleF . toTuple
+
+instance Buildable (ForResponseLog Params4PL) where
+  build = buildForResponse
 
 -- Serializing as a tuple for brevity.
 instance ToJSON Params4PL where
--- a/backend/src/Edna/Dashboard/DB/Query.hs
+++ b/backend/src/Edna/Dashboard/DB/Query.hs
@@ -5,7 +5,6 @@ module Edna.Dashboard.DB.Query
   , setNameSubExperiment
   , setIsSuspiciousSubExperiment
   , deleteSubExperiment
-  , createSubExperiment
   , getExperiments
   , getDescriptionAndMetadata
   , getFileNameAndBlob
@@ -24,14 +23,11 @@ import Database.Beam.Backend (SqlSerial(..))
 import Database.Beam.Postgres (PgJSON(..), Postgres)
 import Database.Beam.Postgres.Full (deleteReturning)
 import Database.Beam.Query
-  (Q, QExpr, aggregate_, all_, as_, asc_, cast_, countAll_, default_, guard_, insert,
-  insertExpressions, int, leftJoin_, lookup_, orderBy_, select, subquery_, update, val_, (&&.),
-  (<-.), (==.))
+  (Q, QExpr, aggregate_, all_, as_, asc_, cast_, countAll_, guard_, int, leftJoin_, lookup_,
+  orderBy_, select, subquery_, update, val_, (&&.), (<-.), (==.))
 
-import Edna.Analysis.FourPL (Params4PL)
 import Edna.DB.Integration
-  (runDeleteReturningList', runInsertReturningOne', runSelectReturningList', runSelectReturningOne',
-  runUpdate')
+  (runDeleteReturningList', runSelectReturningList', runSelectReturningOne', runUpdate')
 import Edna.DB.Schema (EdnaSchema(..), ednaSchema)
 import Edna.Dashboard.DB.Schema
 import Edna.Dashboard.Web.Types (ExperimentResp(..))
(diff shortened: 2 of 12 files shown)


backend/src/Edna/Dashboard/DB/Query.hs
```diff
@@ -156,3 +205,12 @@ getRemovedMeasurements (SqlId subExperimentId) =
...
     pure removedMeasurement
+
+-- Helper
+getExperimentId :: SubExperimentId -> Edna Word32
+getExperimentId sqlId@(SqlId subExpId) = do
+  mRes <- runSelectReturningOne' $ select $ do
+    subExperiment <- all_ $ esSubExperiment ednaSchema
+    guard_ (seSubExperimentId subExperiment ==. val_ (SqlSerial subExpId))
+    pure (seExperimentId subExperiment)
+  justOrThrow (DESubExperimentNotFound sqlId) mRes
```
sashasashasasha151: So, again, it looks more convenient to throw errors from Service part and here return Maybe Word32

backend/src/Edna/Dashboard/DB/Query.hs
```diff
@@ -77,6 +86,24 @@ deleteSubExperiment (SqlId subExpId) =
      )
   ) seSubExperimentId
 
+-- | Create a new sub-experiment as a sibling of existing one (sharing
+-- the same experiment ID).
+cloneSubExperiment :: SubExperimentId -> Text -> Params4PL -> Edna SubExperimentRec
+cloneSubExperiment sqlId newName result = do
+  experimentId <- getExperimentId sqlId
```
sashasashasasha151: So, here and in `getMeasurements` I think `experimentId` should a function parameter.

backend/src/Edna/Dashboard/Service.hs
```diff
@@ -64,10 +73,53 @@ setIsSuspiciousSubExperiment subExpId isSuspicious = do
 -- primary.
 deleteSubExperiment :: SubExperimentId -> Edna NoContent
 deleteSubExperiment subExpId = NoContent <$ do
+  -- Check existence (not the optimal way, but it likely doesn't matter)
+  void $ getMeasurements subExpId
```
sashasashasasha151: Not really understand what for you need it

backend/src/Edna/Analysis/FourPL.hs
```diff
@@ -49,4 +53,4 @@ instance ToSchema Params4PL where
 --
 -- TODO [EDNA-71] Implement!
 analyse4PL :: [(Double, Double)] -> IO Params4PL
-analyse4PL _ = pure (Params4PL 1 2 3 4)
+analyse4PL _ = pure (Params4PL 36404 1.14 33 -2552)
```
sashasashasasha151, line 56: So, it will be useful but only for one experiment, but ok)

backend/src/Edna/Dashboard/Service.hs
```diff
@@ -64,10 +73,53 @@ setIsSuspiciousSubExperiment subExpId isSuspicious = do
...
 
+-- | Create a new sub-experiment based on existing one.
+newSubExperiment ::
+  SubExperimentId -> NewSubExperimentReq -> Edna (WithId 'SubExperimentId SubExperimentResp)
+newSubExperiment subExpId req = do
+  (removed, newResult) <- analyseNewSubExperiment subExpId req
+  transact $ do
+    result <- subExperimentRecToResp <$>
+      Q.cloneSubExperiment subExpId (nserName req) newResult
+    result <$ insertRemovedMeasurements (wiId result) removed
```
sashasashasasha151: Looks like transact is not needed

sashasashasasha151 (changes requested):

backend/src/Edna/Dashboard/Service.hs
gromakovsky: Without this check we fail with `DECantDeletePrimary` for non-existing sub-experiments.

image

backend/src/Edna/Analysis/FourPL.hs
gromakovsky: According to @pva701 
> Thx, mocks work for me. Will try

So I think it's ok indeed.

backend/src/Edna/Dashboard/Service.hs
gromakovsky: I think we should `cloneSubExperiment` and `insertRemovedMeasurements` atomically. If we insert sub-experiment and don't insert removed measurements (e. g. because the app crashes), we will get invalid data in DB.

From my experiments `transact` doesn't work like that, but it's a topic of another issue (EDNA-76). AFAIU it is supposed to give us atomicity, but doesn't give for some reason. Anyway, even if `transact` does nothing, it's a marker saying "these things should happen atomically".

backend/src/Edna/Dashboard/DB/Query.hs
gromakovsky: Ok, I'll try, doesn't sound hard, but let me see.

backend/src/Edna/Dashboard/Web/API.hs
```diff
@@ -71,6 +94,23 @@ data DashboardEndpoints route = DashboardEndpoints
...
+      :> Get '[JSON] ExperimentMetadata
+
+  , -- | Download experiment data file that stores experiment with given ID
+    deGetExperimentFile :: route
+      :- "experiment"
+      :> Summary "Download experiment data file that stores experiment with given ID"
+      :> Capture "experimentId" ExperimentId
+      :> "file"
+      :> Get '[OctetStream]
+        (Headers '[Header "Content-Disposition" Text] ExperimentFileBlob)
```
sashasashasasha151, line 112: Is it possible somewhere to get file name and description? But as I see there is no such information in design

sashasashasasha151 (changes requested):
Doesn't getSubExperiment should contain some flag about its primary? Also as I understand it is not possible to fetch all sub-experiments for experiment

gromakovsky:
> Doesn't getSubExperiment should contain some flag about its primary? 

This information is part of `ExperimentResp` and @pva701 seems to be ok with that.

> Also as I understand it is not possible to fetch all sub-experiments for experiment

Yes, @pva701 already wrote about that. See [here](https://serokell.slack.com/archives/C01F8AQSHGE/p1617727367281100?thread_ts=1617726546.279700&cid=C01F8AQSHGE). I was told that in practice the number of sub-experiments will most often be 2, 1 will be less often, then 3. More than 3 is a rare case. And also :are_you_serious: started arguing that making queries one-by-one is ok. So, taking all of that into account, I decided to zabeetb.

backend/src/Edna/Dashboard/Web/API.hs
gromakovsky: Yes, if you expand the table in Dashboard
image

`Metadata` shows description and metadata from `.xlsx`, `Download` actually downloads `.xlsx` file.

backend/src/Edna/Dashboard/Service.hs
sashasashasasha151: Them looks like you again need transact here

backend/src/Edna/Dashboard/Web/API.hs
sashasashasasha151: And file name?

sashasashasasha151 (approved):

backend/src/Edna/Dashboard/Web/API.hs
gromakovsky: I put file name into `Content-Disposition` header, so that when you click "Download" this file name will be used by default in the file save dialog.

backend/src/Edna/Dashboard/Service.hs
gromakovsky: Ok, I'll add it.

PR #54, opened 2021-04-09
[EDNA-72] Update Library backend

gromakovsky:
## Description

Problem: for each methodology we display a list of projects that
use this methodology. However, currently backend doesn't return this
data.

Solution: return projects, update frontend. Also update 'last update' for projects when experiments are added.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-72

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/backend/src/Edna/Library/DB/Query.hs
+++ b/backend/src/Edna/Library/DB/Query.hs
@@ -25,6 +25,7 @@ module Edna.Library.DB.Query
 import Universum
 
 import qualified Data.List as L
+import qualified Data.List.NonEmpty as NE
 import qualified Database.Beam.Postgres.Full as Pg
 
 import Database.Beam.Backend (SqlSerial(..))
@@ -41,8 +42,7 @@ import Edna.Dashboard.DB.Schema (ExperimentT(..))
 import Edna.Library.DB.Schema as LDB
   (CompoundRec, CompoundT(..), PrimaryKey(..), ProjectRec, ProjectT(..), TargetRec, TargetT(..),
   TestMethodologyRec, TestMethodologyT(..))
-import Edna.Library.Web.Types
-  (MethodologyReqResp(..), ProjectReq(..), ProjectResp(..), TargetResp(..))
+import Edna.Library.Web.Types (MethodologyReq(..), ProjectReq(..), ProjectResp(..), TargetResp(..))
 import Edna.Setup (Edna)
 import Edna.Upload.DB.Schema (ExperimentFileT(..))
 import Edna.Util as U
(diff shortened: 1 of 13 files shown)


backend/src/Edna/Library/DB/Query.hs
```diff
@@ -178,30 +178,49 @@ getMethodologyByName name = runSelectReturningOne' $ select $ do
...
-getMethodologies :: Edna [TestMethodologyRec]
-getMethodologies = runSelectReturningList' $ select $ all_ $ esTestMethodology ednaSchema
+getMethodologies :: Edna [(TestMethodologyRec, [Text])]
+getMethodologies = getMethodology' Nothing
+
+getMethodology' :: Maybe MethodologyId -> Edna [(TestMethodologyRec, [Text])]
+getMethodology' mMethodologyId =
+  fmap convert $
+  runSelectReturningList' $ select $
+  pgNubBy_ (bimap tmTestMethodologyId pProjectId) $ do
```
sashasashasasha151, line 188: I think we also need `orderBy_`

backend/src/Edna/Library/DB/Query.hs
```diff
@@ -178,30 +178,49 @@ getMethodologyByName name = runSelectReturningOne' $ select $ do
   pure methodologies
 
 -- | Get all methodologies
-getMethodologies :: Edna [TestMethodologyRec]
-getMethodologies = runSelectReturningList' $ select $ all_ $ esTestMethodology ednaSchema
+getMethodologies :: Edna [(TestMethodologyRec, [Text])]
+getMethodologies = getMethodology' Nothing
+
+getMethodology' :: Maybe MethodologyId -> Edna [(TestMethodologyRec, [Text])]
```
sashasashasasha151, line 184: Why not to return `[WithId 'MethodologyId MethodologyResp]`

backend/src/Edna/Library/Web/Types.hs
```diff
@@ -61,25 +63,52 @@ instance ToSchema ProjectResp where
...
+
+instance Buildable (ForResponseLog MethodologyResp) where
+  build = buildForResponse
+
+deriveJSON ednaAesonWebOptions ''MethodologyResp
+
+instance ToSchema URI => ToSchema MethodologyResp where
+  declareNamedSchema = gDeclareNamedSchema
+
+methodologyReqToResp :: MethodologyReq -> [Text] -> MethodologyResp
```
sashasashasasha151: As I can see it only used in tests

sashasashasasha151 (changes requested):

backend/src/Edna/Library/DB/Query.hs
gromakovsky: No, because I use `NE.groupAllWith` which doesn't require sorting.

backend/src/Edna/Library/DB/Query.hs
gromakovsky: I think that using API types in Service modules is something undesirable. I don't want to prohibit it because sometimes it's convenient, but in general I think that conversion between DB types and API types should be the goal of Service layer.

backend/src/Edna/Library/Web/Types.hs
gromakovsky: Yes, and it's used in more than one test module. So it's not obvious where to put it. Do you propose to create `Test.Util`?

gromakovsky: So if you tell me where you think I should put this function, I'll move it there :stuck_out_tongue_winking_eye:

sashasashasasha151: So, `Test.Util` sounds better for me. I just was confused what a strange function I see and found that it is used only in tests)

sashasashasasha151 (approved):

PR #57, opened 2021-04-12
[EDNA-44] Improve config handling

worm2fed:
## Description

Add functionality to pass command line arguments and use environment variables. The most priority has cla, then env and then config.
Also there are changes in docker-compose.yaml, so now we can use .env file to specify db port, user, password and db name. But if we do we also have to export this variables to override default one in backend configuration.

## Related issue(s)

Resolves https://issues.serokell.io/issue/EDNA-44

- Tests
        silently reappearing again.

- Documentation
    - [README](../tree/master/backend/README.md)


.gitignore
```diff
@@ -1 +1,3 @@
 # See https://www.notion.so/serokell/gitignore-3d45b853c4494fa282dc04269f851b2e
+.vscode/
```
gromakovsky: As the link above says, do not:
> Include patterns for temporary or state files created by tools that you use, but others might not. A common example of such a tool is a text editor – there are many of them.

I think ` .vscode/` and ` *.code-workspace` fall into this category. It's better to put them into your global gitignore (the link above says how to do it).

backend/src/Edna/Config/Environment.hs
```diff
@@ -0,0 +1,86 @@
...
+apiListenEnv :: IO (Maybe NetworkAddress)
+apiListenEnv = do
+  env <- lookupEnv "EDNA_API_LISTEN"
+  return $ read <$> env
+
+serveDocsEnv :: IO (Maybe Bool)
+serveDocsEnv = lookupEnv "EDNA_SERVE_DOCS" <&> \case
+  Just "1"                       -> Just True
+  Just (map C.toLower -> "true") -> Just True
+  _                              -> Nothing
```
gromakovsky: It duplicates `askDebugDB`, can you please extract this logic into a separate helper function?

gromakovsky:
image

Stylistic comment: we wrap commit message header to be ≤ 50 characters long (soft limit) and subject line to fit into 72 characters. 72 is also hard limit for the header.

backend/app/Main.hs
```diff
@@ -4,20 +4,16 @@ module Main
...
-import Edna.Setup (runEdna)
+import Edna.Config.CLA (ednaOpts)
+import Edna.Setup (prepareConfig, runEdna)
 import Edna.Web.Server (edna)
 
 main :: IO ()
 main = withUtf8 $ do
+  options <- execParser ednaOpts
+  config <- prepareConfig options
   hPutStrLn @Text stderr "Edna server is started"
```
gromakovsky: I think it's better to put it before options parsing. This message is helpful for debugging (e. g. to debug Docker stuff) when you want to know whether the server was started at all. If we write it after parsing, there is a small chance that the application will die during parsing and the message won't be printed.

backend/README.md
```diff
@@ -20,5 +20,28 @@ Run `stack exec -- edna-server -c dev-config.yaml` or `cabal run -- edna-server
 * The server endpoints will be available at `http://localhost:9000/api/*`.
 * Swagger docs will be available at `http://localhost:9000/docs/`.
 
-Set `EDNA_DEBUG_DB=1` to enable logging of all DB actions made.
+Set `EDNA_DB_DEBUG=1` to enable logging of all DB actions made.
 It works in tests as well.
+
+Also you can pass other command line arguments, to see full list run `stack exec -- edna-server -h` or `cabal run -- edna-server -h`:
+```
+Usage: edna-server [-c|--config EDNA_CONFIG] [--listen-addr EDNA_API_LISTEN]
```
gromakovsky: I slightly dislike having the full help message in README because we'll most likely forget to update it when we add a new option or change existing ones. Without automated checks it's almost impossible to keep it up-to-date. I think just mentioning `edna-server -h` (as you do) is already sufficient.

backend/src/Edna/Config/Environment.hs
```diff
@@ -0,0 +1,86 @@
...
+  , eeDocs = serveDocsEnv
+  , eeConnection = connectionEnv
+  , eeMaxCon = dbMaxConnectionsEnv
+  , eeLogging = loggingLevelEnv
+  }
+
+apiListenEnv :: IO (Maybe NetworkAddress)
+apiListenEnv = do
+  env <- lookupEnv "EDNA_API_LISTEN"
+  return $ read <$> env
```
gromakovsky: `read` is not good because it fails using `error` on bad input. We use `error` only for programmer mistakes, errors that are supposed to never happen no matter what input is provided. If `read` fails, the error may be triggered at some unexpected moment when `NetworkAddress` will be forced. There is `readEither` in `Universum`, it's better to use it here and throw some exception here when it fails (for example you can define `EnvParseError` type or use `throwString` from `safe-exceptions`).

gromakovsky (changes requested):
I have some initial feedback. I want to sleep now, so this review wasn't very attentive, probably I'll have more comments when I come back.

backend/src/Edna/Config/Environment.hs
worm2fed: sure

backend/app/Main.hs
worm2fed: the reason I've put it there - if we run Edna with `-h` option (to see help) we also will see message that Edna was started.. this is not correct

But we can put it one line upper, after options parsing and before configuration. What do you think?

gromakovsky: Okay, sounds like a good compromise, let's do it.

backend/src/Edna/Config/Environment.hs
worm2fed: do you think here is better to throw exception or just put some notice message in IO and assume env is not set?

gromakovsky: I think it's better to throw an exception because "some notice message" is easier to accidentally miss. So if someone sets `EDNA_API_LISTEN` to something, I think they certainly intend to set listen address to something and if it doesn't work they want to restart properly anyway, so it's better to stop the program because it will be stopped anyway.

backend/README.md
gromakovsky: Why are these conversations marked as resolved if they are not? I mean, there are no requested changes (maybe you forgot to push?). And usually the person who requested changes marks a conversation as resolved after checking that changes have been made properly.

deployment/.gitignore
```diff
@@ -0,0 +1,2 @@
+postgres/
```
gromakovsky: So I've just merged #40 and it brought some conflicts. Besides, due to https://github.com/serokell/edna/commit/f65cd28c1d042a0eb34e22802b876761abb4578c we don't create `postgres/` directory here anymore, so you can drop this line.

deployment/docker-compose.yml
```diff
@@ -3,14 +3,16 @@ services:
   postgres:
     image: "postgres"
     container_name: postgres
+    env_file:
+      - .env
     ports:
-      - "5432:5432"
+      - "${POSTGRES_PORT}:5432"
```
gromakovsky: Don't we want to map `POSTGRES_PORT` to `POSTGRES_PORT`?

deployment/docker-compose.yml
```diff
@@ -3,14 +3,16 @@ services:
   postgres:
     image: "postgres"
     container_name: postgres
+    env_file:
+      - .env
```
gromakovsky: Is it correct that in order to use this `docker-compose.yml` now one needs to create `.env` file here? I think it's acceptable, but it needs to be mentioned in documentation (after you rebase this PR).

deployment/docker-compose.yml
worm2fed: `5432` is default postgres port and if we want to change it we have to provide some changes to confirm files (for example). But actually we do not interested in it because we have postgres inside docker as separate service.
So, if somebody wants to not use default postgres port outside of postgres service he can use `POSTGRES_PORT` (or leave this env with default value)

deployment/docker-compose.yml
worm2fed: Yes, we have to use `.env` for now.
What about our test server? I guess we have to add some variables

backend/src/Edna/Config/Environment.hs
worm2fed: agree

backend/README.md
```diff
@@ -26,11 +26,17 @@ If you use the config from this repository:
 * Swagger docs will be available at `http://localhost:9000/docs/`.
 
 Note that you need to also run PostgreSQL server to make backend work.
-One way to run it is to use `docker-compose.yml` provided in the [deployment](../deployment) folder.
+One way to run it is to use `docker-compose.yml` provided in the [deployment](../deployment) folder (do not forget to set environment variables related to postgres - create `.env` file from example file [.env.example](../deployment/.env.example))
```
gromakovsky: > (do not forget to set environment variables related to postgres - create `.env` file from example file [.env.example](../deployment/.env.example))

I think it should be mentioned in `deployment/README.md`, either instead of this place or in addition to this place. Because people may go to `deployment` not from `backend` but from root.

backend/dev-config.yaml
```diff
@@ -6,7 +6,7 @@ api:
 db:
   # Syntax: https://www.postgresql.org/docs/9.5/libpq-connect.html#LIBPQ-CONNSTRING
   # It should match parameters from `/deployment/docker-compose.yml`
-  conn-string: "host=localhost dbname=ednadb user=ednauser password=ednapass"
+  conn-string: "host=localhost port=5432 dbname=ednadb user=ednauser password=ednapass"
```
gromakovsky, line 12: AFAIU `port=5432` is the default port, so is it necessary?

backend/src/Edna/Config/CLA.hs
```diff
@@ -0,0 +1,66 @@
...
+  help "Whether to serve docs."
+
+maxConnectionsParser :: Parser (Maybe Int)
+maxConnectionsParser = optional $ option auto $
+  long "max-connections" <>
+  metavar "EDNA_DB_MAX_CONNECTIONS" <>
+  help "Edna DB max connections."
+
+loggingParser :: Parser (Maybe LoggingConfig)
+loggingParser = optional $ option auto $
```
gromakovsky: AFAIU `auto` will parse `LogDev`, `LogProd`, etc., but we need to parse `Dev`, `Prod` and `Nothing` (without `Log` prefix). You can use `eitherReader` here and pass a custom parsing function to it.

backend/src/Edna/Config/Environment.hs
```diff
@@ -0,0 +1,104 @@
...
+
+-- If @EDNA_DB_DEBUG@ environment variable is set to @1@
+-- or @TRUE@ (case-insensitive), we will run postgres with
+-- debug logging.
+askDebugDB :: IO Bool
+askDebugDB = do
+  env <- getBoolEnv "EDNA_DB_DEBUG"
+  return $ fromMaybe False env
+
+data EdnaEnvironments = EdnaEnvironments
```
gromakovsky: Do we need this type and to pass a value of this type in some functions? I mean, all `IO` actions in this type are global functions that are available everywhere, so instead of carrying `ednaEnvs` and using `eeListen ednaEnvs` we can just use `apiListenEnv` directly. For example, in `prepareConfig` you do `let envOptions = ednaEnvs` and pass `envOptions` further, but in that place you can also access `ednaEnvs`.

backend/src/Edna/Config/Environment.hs
```diff
@@ -0,0 +1,104 @@
...
+
+connectionEnv :: IO (Maybe ConnString)
+connectionEnv = do
+  host <- lookupEnv "POSTGRES_HOST"
+  port <- lookupEnv "POSTGRES_PORT"
+  user <- lookupEnv "POSTGRES_USER"
+  password <- lookupEnv "POSTGRES_PASSWORD"
+  db <- lookupEnv "POSTGRES_DB"
+
+  if any isNothing [host, port, user, password, db]
```
gromakovsky: I couldn't find it in docs, but AFAIU if one keyword appears in `conn-string`, the last one takes precedence. So for example the following works fine: `"host=1.2.3.4 host=localhost dbname=ednadb user=ednauser password=ednapass"`. For this reason I think it's not necessary to require all variables to be defined. For example, if only `POSTGRES_HOST` is defined, we can append `host=$POSTGRES_HOST` to the connection string and it should work fine (override the host if it was provided in the config).

deployment/docker-compose.yml
gromakovsky: Do you mean [staging](https://staging.edna.serokell.team/) server?  Well, we can update its config, but I think it's not necessary because the changes in this PR are backwards compatible (correct me if I am wrong).

gromakovsky (commented):
I'd like to see some tests also. In order to do it I propose to add `dump-config` command which prints final config in yaml format. By "final" I mean that all command line options and environment variables are taken into account. Tests should run `edna-server` executable. We typically use [`bats`](https://github.com/sstephenson/bats) for such tests. You will need to add the corresponding step to CI config (`.buildkite/pipeline.yml`). To make `bats` available, you need to run it inside `nix-shell`, like we run `hlint` for example. `nix-shell` environment is specified in `flake.nix` (see `devShell` there).

deployment/docker-compose.yml
worm2fed: If staging server uses postgres in docker - then we have to path `.env` file on build stage with the same variables as we use in our config file for edna

backend/dev-config.yaml
worm2fed: no, it is not, but for now we have ability to set another db port in env vars, so I guess it is useful to show such an option in regular config file

backend/src/Edna/Config/Environment.hs
gromakovsky: Let me repeat myself:
> Why are these conversations marked as resolved if they are not? I mean, there are no requested changes (maybe you forgot to push?). And usually the person who requested changes marks a conversation as resolved after checking that changes have been made properly.

worm2fed: No, I do not forgot to push, just wanted to make some more changes in commit.
I mark it as resolved because I agree with you and made changes locally. So it's just easy to read other comments in PR.
If it is a problem, I will not close it)

backend/src/Edna/Config/Environment.hs
worm2fed: Yes, you are right. I've tried `"host=1.2.3.4 port=5432 dbname=test user=ednauser password=ednapass host=localhost dbname=ednadb"` and it's working.

backend/src/Edna/Config/Environment.hs
gromakovsky: The problem with resolving before the reviewer checked the change is that it makes it harder for review to re-review PR after some changes have been made. Normally I (or another reviewer) write some comments, you make changes, I come to the PR again and go through my comments checking which of them have been fixed in some way (by making changes in code or explaining why no changes are needed or by some other means). If my comments are marked as resolved, I need to unfold each of them to recall what I was requesting.
It's ok to resolve comments where no changes are requested, e. g. sometimes I can write "I like how it's done" and you can resolve it as soon as you read it.

> So it's just easy to read other comments in PR.

For this purpose I suggest that you just write "Done" or "Ok" for comments that you already took into account. Then you can go through all comments and pay attention only to those not ending with "Done" or "Ok" from your side.

backend/dev-config.yaml
gromakovsky: Okay, I don't mind actually.

deployment/docker-compose.yml
gromakovsky: Staging server runs Docker images without `docker-compose` and it doesn't use our `deployment` folder at all. It runs them using `nix` somehow (I don't know how exactly). It only provides config file in the right place: https://github.com/serokell/gemini-infra/blob/62b9bd5988ba9a91e6d56c3ca4c156708b2e7697/servers/castor/edna.nix

And this approach should still work after this PR.

backend/scripts/test.bats
```diff
@@ -0,0 +1,108 @@
...
+
+
+@test "Passing wrong EDNA_LOGGING_LEVEL cla causes error" {
+  run edna-server --logging wrong_logging_level
+  assert_failure
+  assert_output --partial "EDNA_LOGGING_LEVEL"
+}
+
+
+@test "Passing clas modifes config" {
```
gromakovsky: It took me a while to understand whether `clas` is `class` with forgotten `s` or something else :) Let's write it as "CLAs".

backend/src/Edna/Setup.hs
```diff
@@ -10,14 +13,18 @@ module Edna.Setup
 
 import Universum
 
-import qualified Data.Char as C
-
-import Lens.Micro.Platform (makeLenses)
+import Data.ByteString as B
```
gromakovsky: This import brings the whole `Data.ByteString` into scope, it's prohibited by our style guide, i. e. all imports from external packages should be explicit. Perhaps you intended to make it `qualified`.

```suggestion
import qualified Data.ByteString as B
```

backend/src/Edna/Setup.hs
```diff
@@ -41,11 +51,51 @@ runEdna config action = do
...
+  listen <- chooseConfig (_acListenAddr apiConfig) (eoListen claOptions) apiListenEnv
+  docs <- chooseConfig (_acServeDocs apiConfig) (eoDocs claOptions) serveDocsEnv
+  return $ ApiConfig listen docs
+
+prepareConnection :: ConnString -> IO ConnString
+prepareConnection conn = do
+  envConn <- connectionEnv
+  let connection = case envConn of
+                    Nothing -> conn
+                    Just c -> ConnString $ unConnString conn <> unConnString c
```
gromakovsky: 1. Shouldn't we insert `" "` between `ConnString`s?
2. I guess you can define `Semigroup` and `Monoid` instances for `ConnString`, they seem reasonable to me.

backend/src/Edna/Config/Definition.hs
```diff
@@ -60,7 +60,7 @@ data LoggingConfig =
   -- ^ Production logging mode: log less than in development mode.
   | LogNothing
   -- ^ No logging.
-  deriving stock (Generic, Show, Eq)
+  deriving stock (Generic, Read, Show, Eq)
```
gromakovsky: Probably you don't need `Read` anymore.

backend/src/Edna/Config/CLA.hs
```diff
@@ -0,0 +1,81 @@
...
+  short 'l' <>
+  long "logging" <>
+  metavar "EDNA_LOGGING_LEVEL" <>
+  help "Edna logging level: Dev, Prod or Nothing."
+
+loggingParser' :: String -> Either String LoggingConfig
+loggingParser' p = case map C.toLower p of
+  "dev"     -> Right LogDev
+  "prod"    -> Right LogProd
+  "nothing" -> Right LogNothing
```
gromakovsky: I propose to put this function next to `LoggingConfig` definition and use it in 3 places:
1. `FromJSON` instance.
2. `loggingParser`
3. `loggingLevelEnv`

backend/scripts/test.bats
```diff
@@ -0,0 +1,108 @@
...
+  run edna-server --listen-addr "127.0.0.1:80" --serve-docs --max-connections 1 --logging prod --dump-config
+  assert_success
+  assert_line --index 1 "api:"
+  assert_line --index 2 "  listen-addr: 127.0.0.1:80"
+  assert_line --index 3 "  serve-docs: true"
+  assert_line --index 4 "logging: LogProd"
```
gromakovsky: Actually it's pretty weird that we parse `Prod` but print `LogProd` (with `Log` prefix). Apparently there is a bug in
```
instance Aeson.ToJSON LoggingConfig where
  toEncoding = Aeson.genericToEncoding loggingConfigAesonOptions
```

it should define not only `toEncoding` but also `toJSON`.

backend/src/Edna/Config/Environment.hs
worm2fed: Ok)

backend/src/Edna/Setup.hs
worm2fed: yes, you are right! I wanted to add qualified import

backend/src/Edna/Config/Definition.hs
worm2fed: done

backend/src/Edna/Setup.hs
worm2fed: I put a space when build hole ConnString from envs
But it'll be better to add instances

backend/src/Edna/Config/CLA.hs
worm2fed: done
but I've put parser function to Definition file because of circle dependencies

backend/scripts/test.bats
worm2fed: done

backend/README.md
worm2fed: done

backend/src/Edna/Config/CLA.hs
worm2fed: done

backend/src/Edna/Config/Definition.hs
```diff
@@ -62,17 +64,24 @@ data LoggingConfig =
...
-  , Aeson.constructorTagModifier = drop 3
   }
 
-instance Aeson.FromJSON LoggingConfig where
-  parseJSON = Aeson.genericParseJSON loggingConfigAesonOptions
+instance Aeson.FromJSON LoggingConfig
 
 instance Aeson.ToJSON LoggingConfig where
   toEncoding = Aeson.genericToEncoding loggingConfigAesonOptions
+  toJSON = Aeson.toJSON . T.toLower . T.drop 3 . show
```
gromakovsky: Erm, it looks wrong.
1. Why did you remove `Aeson.constructorTagModifier` from `loggingConfigAesonOptions`? It used to drop `Log` prefix, but now it doesn't.
2. `toEncoding` and `toJSON` should have identical definitions, but now the are different. `toJSON` drops the prefix, but `toEncoding` doesn't. You can define `toJSON = Aeson.genericToJSON loggingConfigAesonOptions` to make it identical.
3. ` instance Aeson.FromJSON LoggingConfig` with no explicit definition will likely use default parsing options which I guess do not drop prefix.  Since we have `parseLoggingConfig` now, I think it's better to define `parseJSON` in terms of `parseLoggingConfig`.

worm2fed: Fixed.
I'll read a bit more about Aeson on weekends)

gromakovsky:
Looks ok now, only need to make CI pass (which is non-trivial).

worm2fed:
@gromakovsky Check please, I've cleaned up my commits

backend/src/Edna/Config/Environment.hs
```diff
@@ -0,0 +1,97 @@
...
+  return $ fromMaybe False env
+
+apiListenEnv :: IO (Maybe NetworkAddress)
+apiListenEnv = readEnv "EDNA_API_LISTEN" WrongNetworkAddress
+
+serveDocsEnv :: IO (Maybe Bool)
+serveDocsEnv = getBoolEnv "EDNA_SERVE_DOCS"
+
+buildConnectionOption :: ByteString -> Maybe String -> Maybe ByteString
+buildConnectionOption option env = (<>) <$> Just option <*> (pack <$> env)
```
gromakovsky: 1. I think it can be simplified as follows: `(option <>) <$> (pack <$> env)` or even `(option <>) . pack <$> env`.
2. `Data.ByteString.Char8.pack` is dangerous because it truncates all characters to 8 bits, it's very rarely a good idea to use it. I think you need `encodeUtf8` here.

backend/src/Edna/Config/CLA.hs
```diff
@@ -0,0 +1,75 @@
...
+
+
+ednaOpts' :: Parser EdnaOptions
+ednaOpts' = EdnaOptions
+        <$> configPathParser
+        <*> listenParser
+        <*> docsParser
+        <*> maxConnectionsParser
+        <*> loggingParser
+        <*> dumpConfigParser
```
gromakovsky, line 46: Where are DB initialization options? Actually the only reason why I sometimes change local config is to erase DB (I set `mode: "enable-with-drop"` for this purpose).

Also AFAIU the only way to override DB connection string is using env variables, which I think is acceptable, but at the same time I think that having a command line option for the whole connection string wouldn't harm.

deployment/.env.example
```diff
@@ -0,0 +1,13 @@
+# database container
+POSTGRES_HOST="localhost"
```
gromakovsky: 1. It should be `postgres` because that's what we have in `config.yaml` in this folder.
2. Regarding " # database container", only `POSTGRES_USER`, `POSTGRES_PASSWORD` and `POSTGRES_DB` options are actually used by `postgresql` image. We also use `POSTGRES_POST` in `ports`. And this HOST variable is just for `edna-server`, so maybe it's more appropriate to call it `EDNA_POSTGRES_HOST` to avoid confusion, but I am not sure.
3. I would add a comment in `deployment/config.yaml` reminding that all options can be overridden by env variables from env file. Also it may serve as a reminder that a new variable should be added when a new configuration field is added.
4. AFAIU the reason why options like `POSTGRES_PASSWORD` are read by `edna-server` is to avoid duplicating them in `edna-server` and `postgres` configuration. But currently they are duplicated which defeats the idea. I think `conn-string` in `deployment/config.yaml` should be empty because the final connection string will be built from these env variables.

backend/src/Edna/Util.hs
```diff
@@ -68,6 +68,12 @@ instance ToJSON ConnString where
 instance Buildable ConnString where
   build (ConnString s) = build $ decodeUtf8 @Text s
 
+instance Semigroup ConnString where
+  ConnString a <> ConnString b = ConnString $ a <> " " <> b
```
gromakovsky, line 85: Oh, this definition violates `Monoid` laws because `mempty <> mempty` differs from `mempty`. The right definition should insert space only if both `a` and `b` are not empty.

deployment/docker-compose.yml
```diff
@@ -22,12 +22,14 @@ services:
   postgres:
     image: postgres
     container_name: postgres
+    env_file:
+      - .env
     ports:
-      - 5432:5432
+      - ${POSTGRES_PORT}:5432
     environment:
```
gromakovsky: Currently this section is totally redundant. You set `POSTGRES_USER` to the current value of `${POSTGRES_USER}`, it does nothing. Same for 2 other variables.

backend/src/Edna/Config/Environment.hs
worm2fed: thanks
fixed

deployment/docker-compose.yml
```diff
@@ -22,12 +22,14 @@ services:
   postgres:
     image: postgres
     container_name: postgres
+    env_file:
+      - .env
```
gromakovsky: I gave it another thought, because I think we want `docker-compose up` to just work without any manual configuration. Copying `.env.example` into `.env` is easy, but still some manual configuration.

Ideally, it should be necessary to create `.env` file only if you want to customize configuration for your needs. I was thinking about using `.env.local` file for customization, but then we need to add `.env.local` to `env-file` and it won't work for people who don't have `.env.local` file. Then I noticed that `docker-compose` has `--env-file` option. So I think that people who want to customize something can create `.env.local` using `.env.example` as a template and pass `--env-file .env.local`. Then we can rename `.env.example` to just `.env`, keep it in the repo and gitignore `.env.local` instead. What do you think?

gromakovsky (changes requested):

backend/src/Edna/Config/Environment.hs
gromakovsky: And can you please test whether it works if you use special characters (e. g. Russian letters) in password for example? I think the version with `pack` shouldn't work and the version with `encodeUtf8` should work.

backend/src/Edna/Config/CLA.hs
worm2fed: Okay. I'll add both

deployment/.env.example
worm2fed: 1. Yes, you're right.
2. I guess it's better to leave as it is. We also have `POSTGRES_POST` in `docker-compose.yaml`, where `POSTGRES_POST` is exposed port for postgres
3. It will be useful
4. Okay)

deployment/docker-compose.yml
worm2fed: Hm.. I have to check that passing `--env-file` overrides `env_file` in compose file. If it works we can simplify running postgres as you propose

deployment/docker-compose.yml
worm2fed: Can you please explain what do you mean?
I'm a bit confused.
`POSTGRES_USER: ${POSTGRES_USER}`
here `POSTGRES_USER` - is a variable inside container and `${POSTGRES_USER}` - is the value we assign to internal container variable. In our case we specify `POSTGRES_USER` in env file, so to use it in compose file we write `${POSTGRES_USER}`.

backend/src/Edna/Config/Environment.hs
worm2fed: With `POSTGRES_PASSWORD="тест"` I have `password=B5AB` in connection string in case of `pack` version, and the same one with `encodeUtf8` with `--dump-config`
But in repl:
```
Data.ByteString.Char8 Universum
λ> (option <>) . encodeUtf8 <$> env
Just "test\209\130\208\181\209\129\209\130"
it :: Maybe ByteString
(0.02 secs, 1,376,560 bytes)

Data.ByteString.Char8 Universum
λ> (option <>) . pack  <$> env
Just "testB5AB"
it :: Maybe ByteString
(0.01 secs, 1,353,136 bytes)
```

So I the problem is somewhere in dump-config

worm2fed: With `POSTGRES_PASSWORD="тест"` I have `password=B5AB` in connection string in case of `pack` version, and `password= тест` in case of `encodeUtf8` (with `--dump-config`)

deployment/docker-compose.yml
worm2fed: yes, it's working as expected

deployment/docker-compose.yml
gromakovsky: If I remove this section, `docker-compose up postgres` still works, that's how I concluded that this section is redundant. What's the difference between "a variable inside container" and "internal container variable"? AFAIU `env_file` and `environment` are used for essentially the same thing: setting variables inside the container. There seems to be no "internal container variable" or I don't understand what it means. My experiment confirms that.

From https://github.com/compose-spec/compose-spec/blob/master/spec.md:
> `env_file` adds environment variables to the container based on file content.

This description sounds like all variables from `env_file` will be set inside `postgres` container.
I just don't understand what would be the point of `env_file` if it worked differently. If we had to do `VAR: ${VAR}` for each variable, it would be really strange.

backend/dev-config.yaml
```diff
@@ -16,5 +16,5 @@ db:
     mode: "enable"
     init-script: "./sql/init.sql"
 
-# Other possible values: 'Prod' and 'Nothing'.
-logging: Dev
+# Other possible values: "prod" and "nothing".
+logging: "dev"
```
gromakovsky, line 23: Oh, while we are at it, there is an old comment a few lines above:
>   # It should match parameters from `/deployment/docker-compose.yml`

It's no longer the case, now these options don't have to match `/deployment/docker-compose.yml` because we have a completely separate config for `deployment`. Can you please fix it (it's unrelated to your issue, but it's a trivial change, so I think it's ok to do it here)?

backend/src/Edna/Config/Definition.hs
```diff
@@ -87,7 +87,7 @@ instance Aeson.FromJSON LoggingConfig where
     let value = parseLoggingConfig Nothing $ T.unpack text
     in case value of
       Right x -> return x
-      Left _  -> fail "Failed to parse LoggingConfig option"
+      Left _  -> error $ "Invalid config value: " <> show text
```
gromakovsky: `fail` is better. `error` is ok for programmer mistakes – code paths that are expected to never be executed regardless of external input. This `FromJSON` instance is used to parse application config that is provided by a user, so we can't be sure that error will never happen. Users may supply malformed values in the config and `fail` is better in this case, because with `fail` it will be more predictable when the error will happen.

backend/src/Edna/Config/Environment.hs
gromakovsky: > (with `--dump-config`)

Okay, and does it actually connect to DB if you run `docker-compose up postgres` and `edna-server` with `POSTGRES_PASSWORD="тест"`? I. e. did you test it without `--dump-config`? Just in case.

worm2fed: it's working with local postgres and edna but with postgres in docker changing password env doesn't has effect.
This is because at first start postgres creates db and roles and do not change password for role on env change. Simply we can remove volume to recreate it.

btw, about your question about russian letters in password - it's working

backend/src/Edna/Config/Definition.hs
worm2fed: I've found that we have another similar definition with `error`, so I've changed it here to use the same approach. I'll remember that `fail` is better and change it

backend/dev-config.yaml
worm2fed: sure

deployment/docker-compose.yml
worm2fed: oh, you are talking about both passing `env_file` and assigning values 
I've got it

backend/src/Edna/Util.hs
worm2fed: strange... I made changes to this definition, but this code has no mark `outdated`..

deployment/docker-compose.yml
```diff
@@ -27,14 +27,11 @@ services:
   postgres:
     image: postgres
     container_name: postgres
+    restart: always
```
gromakovsky, line 54: Why?

worm2fed: In docker doc we have following notice: `Docker recommends that you use restart policies, and avoid using process managers to start containers.`
The default value is `no`.

>`no` - Do not automatically restart the container. (the default)
>`on-failure` - Restart the container if it exits due to an error, which manifests as a non-zero exit code.
>`always` - Always restart the container if it stops. If it is manually stopped, it is restarted only when Docker daemon restarts or the container itself is manually restarted. (See the second bullet listed in restart policy details)
>`unless-stopped` - Similar to always, except that when the container is stopped (manually or otherwise), it is not restarted even after Docker daemon restarts.
(https://docs.docker.com/config/containers/start-containers-automatically/)

Now imagine that you run Edna somewhere on your server and accidentally your server fall down. You restart server and forgot to manually run your containers. With this setting (if your docker daemon starts automatically) postgres container will start automatically. 

I guess we have to add similar option to other services too.

gromakovsky: Ok, I see that it still doesn't automatically restart if I manually stop the container (e. g. click `Ctrl-C` after `docker-compose up postgres`), so it's fine. This setting doesn't really matter for us because we don't use `docker-compose.yaml` on server, we only use it locally. But perhaps it can be useful for other people who may want to run Edna on their server, so I'm ok with this setting.

Please don't add it _silently_ though, write a sentence or two about why it's done. Ideally it should be put into its own commit because it looks independent of other changes.

deployment/docker-compose.yml
```diff
@@ -11,28 +14,31 @@ services:
...
       - 9000:9000
     command: --config /config.yaml
+
   frontend:
     image: "ghcr.io/serokell/edna-frontend:latest"
     depends_on:
       - backend
     ports:
-      - 8080:80
+      - ${FRONTEND_PORT}:80
```
gromakovsky, line 47: Does it work without `env_file` for frontend? How does it know `FRONTEND_PORT`?

backend/README.md
gromakovsky: Perhaps it can be reverted now, because creating `.env` files is no longer necessary after recent changes.

backend/README.md
```diff
@@ -26,11 +26,17 @@ If you use the config from this repository:
...
 
-Set `EDNA_DEBUG_DB=1` to enable logging of all DB actions made.
+Set `EDNA_DB_DEBUG=1` to enable logging of all DB actions made.
 It works in tests as well.
 
+Also you can pass other command line arguments, to see full list run `stack exec -- edna-server -h` or `cabal run -- edna-server -h`:
+
+Also you can use environment variables to configure your server. You can find supported variables in [`.env.example`](../deployment/.env.example).
```
gromakovsky: The link needs to be updated now.

backend/src/Edna/Util.hs
```diff
@@ -104,13 +119,22 @@ instance ToJSON NetworkAddress where
 ----------------
 
 data DatabaseInitOption = Enable | EnableWithDrop
-  deriving stock (Generic, Show)
+  deriving stock (Generic, Read, Show)
```
gromakovsky: Is `Read` needed?

gromakovsky (changes requested):
Btw, did you test that `docker-compose up` works out of the box with these changes? Note that it pulls latest images from GHCR which are updated on each push to master. And those master images are not compatible with this branch (they don't parse env variable for example and we now rely on them). In order to test it, you can download images from CI artifacts and load them with `docker load -i`, it will force usage of those images instead of fetching them from GHCR.

gromakovsky:
It looks very close to be finished. Btw, you can squash existing fixup commits, I've checked them.

worm2fed:
> Btw, did you test that `docker-compose up` works out of the box with these changes? Note that it pulls latest images from GHCR which are updated on each push to master. And those master images are not compatible with this branch (they don't parse env variable for example and we now rely on them). In order to test it, you can download images from CI artifacts and load them with `docker load -i`, it will force usage of those images instead of fetching them from GHCR.


image
Strange date))) Is time traveling in law?

Yes, it's working

worm2fed:
> It looks very close to be finished. Btw, you can squash existing fixup commits, I've checked them.

done

deployment/config.yaml
```diff
@@ -1,15 +1,14 @@
 # Config for docker-compose-based local deployment
+# NOTE: All this options can be overwritten be env variables
```
gromakovsky:
```suggestion
# NOTE: All these options can be overwritten by env variables
```

deployment/docker-compose.yml
gromakovsky: @worm2fed ping.

worm2fed: @gromakovsky oh, I thought I answered.
https://github.com/serokell/edna/pull/57#issuecomment-824816992

it's working

worm2fed: image

image

image

gromakovsky:
>  Strange date))) Is time traveling in law?

I think it's a very minor problem, but it's better to provide correct time there. We are using [buildLayeredImage](https://nixos.org/manual/nixpkgs/stable/#ssec-pkgs-dockerTools-buildLayeredImage) and it has `created` option which defaults to `1970-01-01T00:00:01Z`. So basically we need to set it to the right time. Can you create an issue in YT please?

deployment/.env
```diff
@@ -0,0 +1,21 @@
...
+POSTGRES_DB="ednadb"
+
+# backend container
+EDNA_API_LISTEN_ADDR="*:9000"
+EDNA_API_SERVE_DOCS=true
+EDNA_DB_MAX_CONNECTIONS=200
+EDNA_DB_INITIALISATION_MODE="enable"
+# Init script is relative to backend image root
+EDNA_DB_INITIALISATION_INIT_SCRIPT="/init.sql"
+EDNA_DB_DEBUG=true
```
gromakovsky: Previously we didn't enable it, so we shouldn't enable it now as well. Application logging is sufficient to understand what's going on, dumping all DB operations is too verbose.

deployment/docker-compose.yml
gromakovsky: Okay, but how does it know above `.env`? We don't specify `env_file` for `frontend`, only for `backend` and `postgres`.

worm2fed: `.env` is default file
https://docs.docker.com/compose/environment-variables/#the-env-file

deployment/.env
worm2fed: I add it here because we already had this env in our readme
Do you propose to delete this variable at all?

deployment/docker-compose.yml
gromakovsky: Ah, ok, I see. Does it mean that all `env_file` options are not necessary? I think it's ok to have `env_file` even if it's not necessary, but I don't understand why we have `env_file` for `backend` and `postgres`, but not for `frontend`.

deployment/.env
gromakovsky: Not delete, can we do `EDNA_DB_DEBUG=false`?

deployment/docker-compose.yml
worm2fed: Yes, it's not necessary. I didn't know that `.env` is default env file and forgot it to add `env_file` to frontend service)
How will be better, to dele `env-file` or add it to frontend service?

deployment/.env
worm2fed: Oh, you are talking about default value, sure!

deployment/docker-compose.yml
gromakovsky: I would delete it and just write a comment somewhere saying that `.env` is the default `env_file`. But also please check one thing: if we delete `env_file` and specify `--env-file .env.local`, will `.env` still be taken into account?

worm2fed: Oh, look what I've found! `.env` is default, but it's applied only for compose file, and it will no forward environment inside service. So we have to keep `env-file` section in compose file for backend and postgres.

Also! passing `--env-file` seems like changing only compose file variables, so docker will no forward env inside service, same as with default file.

The decision can be specify two files in compose file:
```yaml
env_file:
      - .env
      - .env.local
```
The second file overrides env vars from first one.

What we have to do: is return back `environment` sections (and keep `env-file`), than it will forward variables inside service and allow to override them with `--env-file`


How docker builds config:
Content of `.env.local`:
```env
POSTGRES_PORT=5431
POSTGRES_DB="local-ednadb"

EDNA_API_SERVE_DOCS=false

FRONTEND_PORT=8085
```

Without `env-file` in compose file
```yaml
docker-compose --env-file=.env.local  config                                                                                                                      0 (0.527s) < 16:50:20
networks:
  edna-network: {}
services:
  backend:
    command: --config /config.yaml
    depends_on:
      postgres:
        condition: service_started
    image: ghcr.io/serokell/edna-backend:latest
    networks:
      edna-network: null
    ports:
    - published: 9000
      target: 9000
    restart: always
    volumes:
    - /Users/worm2fed/workspace/serokell/edna/deployment/config.yaml:/config.yaml:rw
  frontend:
    depends_on:
      backend:
        condition: service_started
    image: ghcr.io/serokell/edna-frontend:latest
    networks:
      edna-network: null
    ports:
    - published: 8085
      target: 80
    restart: always
  postgres:
    container_name: postgres
    image: postgres
    networks:
      edna-network: null
    ports:
    - published: 5431
      target: 5432
    restart: always
    volumes:
    - edna-db-data:/var/lib/postgresql/data:rw
version: '3.7'
volumes:
  edna-db-data: {}
 ```
 
 When I put it back:
 ```yaml
 networks:
  edna-network: {}
services:
  backend:
    command: --config /config.yaml
    depends_on:
      postgres:
        condition: service_started
    environment:
      EDNA_API_LISTEN_ADDR: '*:9000'
      EDNA_API_SERVE_DOCS: "true"
      EDNA_DB_DEBUG: "false"
      EDNA_DB_INITIALISATION_INIT_SCRIPT: /init.sql
      EDNA_DB_INITIALISATION_MODE: enable
      EDNA_DB_MAX_CONNECTIONS: '200'
      EDNA_LOGGING: dev
      FRONTEND_PORT: '8080'
      POSTGRES_DB: ednadb
      POSTGRES_HOST: postgres
      POSTGRES_PASSWORD: ednapass
      POSTGRES_PORT: '5432'
      POSTGRES_USER: ednauser
    image: ghcr.io/serokell/edna-backend:latest
    networks:
      edna-network: null
    ports:
    - published: 9000
      target: 9000
    restart: always
    volumes:
    - /Users/worm2fed/workspace/serokell/edna/deployment/config.yaml:/config.yaml:rw
  frontend:
    depends_on:
      backend:
        condition: service_started
    image: ghcr.io/serokell/edna-frontend:latest
    networks:
      edna-network: null
    ports:
    - published: 8085
      target: 80
    restart: always
  postgres:
    container_name: postgres
    environment:
      EDNA_API_LISTEN_ADDR: '*:9000'
      EDNA_API_SERVE_DOCS: "true"
      EDNA_DB_DEBUG: "false"
      EDNA_DB_INITIALISATION_INIT_SCRIPT: /init.sql
      EDNA_DB_INITIALISATION_MODE: enable
      EDNA_DB_MAX_CONNECTIONS: '200'
      EDNA_LOGGING: dev
      FRONTEND_PORT: '8080'
      POSTGRES_DB: ednadb
      POSTGRES_HOST: postgres
      POSTGRES_PASSWORD: ednapass
      POSTGRES_PORT: '5432'
      POSTGRES_USER: ednauser
    image: postgres
    networks:
      edna-network: null
    ports:
    - published: 5431
      target: 5432
    restart: always
    volumes:
    - edna-db-data:/var/lib/postgresql/data:rw
version: '3.7'
volumes:
  edna-db-data: {}
```

With `env-file` and `environment` in compose file
```yaml
WARNING: The POSTGRES_HOST variable is not set. Defaulting to a blank string.
WARNING: The POSTGRES_USER variable is not set. Defaulting to a blank string.
WARNING: The POSTGRES_PASSWORD variable is not set. Defaulting to a blank string.
WARNING: The EDNA_API_LISTEN_ADDR variable is not set. Defaulting to a blank string.
WARNING: The EDNA_DB_MAX_CONNECTIONS variable is not set. Defaulting to a blank string.
WARNING: The EDNA_DB_INITIALISATION_MODE variable is not set. Defaulting to a blank string.
WARNING: The EDNA_DB_INITIALISATION_INIT_SCRIPT variable is not set. Defaulting to a blank string.
WARNING: The EDNA_DB_DEBUG variable is not set. Defaulting to a blank string.
WARNING: The EDNA_LOGGING variable is not set. Defaulting to a blank string.
networks:
  edna-network: {}
services:
  backend:
    command: --config /config.yaml
    depends_on:
      postgres:
        condition: service_started
    environment:
      EDNA_API_LISTEN_ADDR: ''
      EDNA_API_SERVE_DOCS: "false"
      EDNA_DB_DEBUG: ''
      EDNA_DB_INITIALISATION_INIT_SCRIPT: ''
      EDNA_DB_INITIALISATION_MODE: ''
      EDNA_DB_MAX_CONNECTIONS: ''
      EDNA_LOGGING: ''
      FRONTEND_PORT: '8080'
      POSTGRES_DB: local-ednadb
      POSTGRES_HOST: ''
      POSTGRES_PASSWORD: ''
      POSTGRES_PORT: '5435'
      POSTGRES_USER: ''
    image: ghcr.io/serokell/edna-backend:latest
    networks:
      edna-network: null
    ports:
    - published: 9000
      target: 9000
    restart: always
    volumes:
    - /Users/worm2fed/workspace/serokell/edna/deployment/config.yaml:/config.yaml:rw
  frontend:
    depends_on:
      backend:
        condition: service_started
    image: ghcr.io/serokell/edna-frontend:latest
    networks:
      edna-network: null
    ports:
    - published: 8085
      target: 80
    restart: always
  postgres:
    container_name: postgres
    environment:
      POSTGRES_DB: local-ednadb
      POSTGRES_HOST: ''
      POSTGRES_PASSWORD: ''
      POSTGRES_PORT: '5435'
      POSTGRES_USER: ''
    image: postgres
    networks:
      edna-network: null
    ports:
    - published: 5435
      target: 5432
    restart: always
    volumes:
    - edna-db-data:/var/lib/postgresql/data:rw
version: '3.7'
volumes:
  edna-db-data: {}
```

worm2fed: so ironic, that my first solution with compose file was correct, but I couldn't explain some points of it)
the good news is that we found a bug and learn more about docker

gromakovsky: Ok, now we are back to my previous question: why do we set `env_file` for `postgres` and `backend`, but don't set it for `frontend`?

gromakovsky: I've pushed one commit with additional comment.

worm2fed: > Ok, now we are back to my previous question: why do we set `env_file` for `postgres` and `backend`, but don't set it for `frontend`?

Because we want to forward variables inside backend and postgres. And we have no env vars inside frontend, so we do not need it.
When (for example) we will want to change listen port of backend and forward it inside frontend image we also will have to add `env_file` section to frontend service (and write some logic to process variable in frontend)

gromakovsky (approved):

PR #58, opened 2021-04-12
[EDNA-78] Add endpoint with metrics for prometheus

worm2fed:
## Description

Problem: We want to know what's going on inside edna-server, at least whether it feels well (in broad sense). We typically use Prometheus for this kind of task.

Solution: Add enpoint with metrics. To collect it we use metrics provided by ghc.

## Related issue(s)

Resolves https://issues.serokell.io/issue/EDNA-78

- Tests
        silently reappearing again.

- Documentation


mkaito (approved):
What exactly do you get from `ghcMetrics`? I assume we'll add relevant domain metrics as we go?

worm2fed:
I get metrics that exposes statistics from the GHC runtime system GHC.Conc (Basic concurrency stuff.) and GHC.Stats (internal garbage collection and memory usage statistics).

Also at this endpoint you can see some requests data, for example:

[metrics.log](https://github.com/serokell/edna/files/6306198/metrics.log)

kirelagin:
> I assume we'll add relevant domain metrics as we go?

That is the plan, but probably not for the MVP.

Regarding `wai-middleware-prometheus`... meh... I guess, that’s good enough for now, but I would really like to see per-endpoint breakdown 🙄. Also this `unsafeRegister` at some arbitrary place in the code makes me wanna scream.

backend/edna.cabal
```diff
@@ -98,6 +98,9 @@ library
     servant-swagger-ui-core,
     servant-swagger-ui,
     servant-util,
+    wai-middleware-prometheus,
+    prometheus-client,
+    prometheus-metrics-ghc,
```
kirelagin: This list is sorted.

backend/edna.cabal
```diff
@@ -115,6 +118,7 @@ executable edna-server
       Paths_edna
   hs-source-dirs:
       app
+  ghc-options: -rtsopts "-with-rtsopts=-T -I0"
```
kirelagin: The `-I0` change does not seem related to this PR, so it would be better to make it separately. But, most importantly, why does it even make sense? Do we have (or even expect to have) any problems with idle GC? If we do, maybe it would be better to just increase the timeout?

backend/src/Edna/Web/API.hs
```diff
@@ -40,7 +40,7 @@ data EdnaEndpoints route = EdnaEndpoints
 type EdnaAPI =
   ToServant EdnaEndpoints AsApi
   :<|>
-  "health" :> Summary "Check the health of this server" :> GetNoContent
+  "metrics" :> "health" :> Summary "Check the health of this server" :> GetNoContent
```
kirelagin: I already commented in Slack, but I’ll repeat here: I find this renaming pretty weird.

backend/stack.yaml
```diff
@@ -36,6 +36,8 @@ extra-deps:
   commit: 3dff2de34cce90c4d1dab68e03d30f8939fd52ca # not master
   subdirs:
     - servant-util
+- prometheus-metrics-ghc-1.0.1.1
+- wai-middleware-prometheus-1.0.0
```
kirelagin: What’s the difference between `wai-middleware-prometheus` and `prometheus-wai-middleware`? I see the latter in Stackage.

backend/README.md
```diff
@@ -22,3 +22,5 @@ Run `stack exec -- edna-server -c dev-config.yaml` or `cabal run -- edna-server
 
 Set `EDNA_DEBUG_DB=1` to enable logging of all DB actions made.
 It works in tests as well.
+
+To collect some data with prometheus you can use `http://localhost:9000/metrics/` endpoint.
```
kirelagin:
```suggestion
To collect some data with prometheus you can use the `http://localhost:9000/metrics/` endpoint.
```

kirelagin (changes requested):

backend/src/Edna/Web/API.hs
worm2fed: @mkaito I guess you can answer here. Does we really need `health` endpoint for now?

backend/edna.cabal
worm2fed: You right, it'll be better to make changes separately.

At first look for now edna is not very complicated and it seems we have no problems with idle. My motivation to add it was that it will not cause problems and this suggestion was provided from relevant area.

If there are objection - it'll be better to remove this change from current PR and discuss it in slack

backend/stack.yaml
worm2fed: The difference is that `prometheus-wai-middleware` depend on [prometheus](http://hackage.haskell.org/package/prometheus) and `wai-middleware-prometheus` depend on [prometheus-client](http://hackage.haskell.org/package/prometheus-client).

worm2fed: here is an example how to use `prometheus-wai-middleware`

```haskell
main :: IO ()
main = runRegistryT $ do
    ms <- applicationMetrics $ fromList [("app", "example_server")]
    let webserver = run 8080 . instrumentApplication ms $ \req respond ->
            respond (makeResponse req)
    _ <- liftIO $ async webserver
    serveMetricsT 8081 []
  where
    success = responseLBS status200 [] "ok"
    failure = responseLBS status404 [] mempty
    makeResponse req
        | pathInfo req == ["test"] = success
        | otherwise = failure
  ```

Comment (worm2fed) [reply backend/stack.yaml]: I took a look at git and `prometheus` has last update two years ago

A think we can stay with `wai-middleware-prometheus` for MVP and further use our `servant-prometheus`.
(be aware, that it also uses `prometheus-client`)

File: backend/src/Edna/Web/API.hs (hunk 3 above)
Comment (mkaito) [reply backend/src/Edna/Web/API.hs]: In a way, he's right. We can remove it. We can just react to the fact that the Prometheus endpoint responds or not.

Comment (worm2fed) [reply backend/src/Edna/Web/API.hs]: done

File: backend/stack.yaml (hunk 4 above)
Comment (gromakovsky) [reply backend/stack.yaml]: Taking all of the above into account, I think using `wai-middleware-prometheus` here is acceptable.

File: backend/src/Edna/Web/API.hs [hunk 6]
```diff
@@ -37,10 +37,7 @@ data EdnaEndpoints route = EdnaEndpoints
   } deriving stock (Generic)
 
 -- | API type specification.
-type EdnaAPI =
-  ToServant EdnaEndpoints AsApi
-  :<|>
-  "health" :> Summary "Check the health of this server" :> GetNoContent
+type EdnaAPI = ToServant EdnaEndpoints AsApi
```
gromakovsky, line 35: Removal of `/health` should be extracted into a separate commit (ideally) or at least be mentioned in the main commit message if you squash everything into one commit (explaining that we can use `/metrics` instead).

backend/src/Edna/Web/Server.hs
```diff
@@ -111,4 +113,7 @@ edna = do
       -- No dev logging, no docs
       | otherwise = serve ednaAPI server
 
-  serveWeb listenAddr loggingConfig app
+  _ <- register ghcMetrics
+  let promMiddleware = prometheus $ PrometheusSettings ["metrics"] True False
```
gromakovsky: Can you please use record syntax to define `PrometheusSettings`? Without explicit record fields it's hard to understand the meaning of 3 values passed to the `PrometheusSettings` constructor.

gromakovsky:
I think it looks ok now besides minor stylistic comments. @mkaito should I create a follow-up issue to connect `/metrics` endpoint with our infrastructure?

backend/src/Edna/Web/API.hs
worm2fed: I'll edit main commit message before merge

worm2fed:
> I think it looks ok now besides minor stylistic comments. @mkaito should I create a follow-up issue to connect `/metrics` endpoint with our infrastructure?

I've finished

backend/src/Edna/Web/API.hs
worm2fed: done

backend/README.md
gromakovsky: Looks like there is no `the` in the end :smile:

gromakovsky (approved):

mkaito:
It's fine for it to be part of the same YT issue. Just assign me and make a comment there.

backend/src/Edna/Web/Server.hs
kirelagin: To be fair, it’s not really because of the record syntax, rather about settings themselves :). With record syntax it would be `prometheusEndPoint`, `prometheusInstrumentApp`, `prometheusInstrumentPrometheus`, which still do not explain anything, IMHO 😆, since they managed to make this configuration type pretty confusing.

I guess what could help a bit is if we used `def` instead to simply hide all this complexity and never think about it again 🙂, and then only use record syntax to override `prometheusInstrumentPrometheus` to `False` (and leave a comment explaining why we do this).

gromakovsky: Well, now it's
```
  let promMiddleware = prometheus $ PrometheusSettings
        { prometheusEndPoint = ["metrics"]
        , prometheusInstrumentApp = True
        , prometheusInstrumentPrometheus = False
        }
```

so I at least more or less understand what `["metrics"]` is.

gromakovsky: > I guess what could help a bit is if we used def instead to simply hide all this complexity and never think about it again slightly_smiling_face, and then only use record syntax to override prometheusInstrumentPrometheus to False (and leave a comment explaining why we do this).

@worm2fed can you do it please?

worm2fed: yes

worm2fed: oh, there are def definition in package already

worm2fed:
@gromakovsky done

gromakovsky:
I've asked Kirill to re-review.

kirelagin (approved):

PR #59, opened 2021-04-13
[EDNA-83] Detect suspicious points in python

HellyHenry:
## Description

Problem: The feature of automatically detecting suspicious data points (a.k.a. outliers).

Solution: Add python script which analyses data points from experiment and makes a decision based on Cook distance of all points. The threshold of selecting outliers is to be chosen later.

## Related issue(s)

Resolves https://issues.serokell.io/issue/EDNA-83

- Tests
        silently reappearing again.

- Documentation


gromakovsky:
If I do `python detect_outliers.py '[{"experiment": 1, "data": [[100, 5973], [33.333, 16689], [11.111, 27737], [3.7037, 33917]]}]'` I get:
> TypeError: Improper input: N=4 must not exceed M=3

Is it expected? What does it mean? I guess, if `detect_outliers.py` exits with non-zero code I will just not add any additional sub-experiment, so it's ok to print anything on bad input, but I still wonder if it's expected in this case.

gromakovsky:
AFAIU in case of bad input `detect_outliers.py` should still succeed and put `"status": "ERROR"` into result, so I guess this error is not expected…

analysis/tests/logic_test.py
```diff
@@ -0,0 +1,43 @@
+import inspect
```
gromakovsky: 1. How should I run this test? I tried `python tests/logic_test.py`, but even if I remove `[150, 1000]` from insane data it returns 0 exit code.
2. I think we want to run this test on CI. @balsoft @mkaito can help with that. AFAIU `analysis/default.nix` should have some `analysis-test` that we'll add to `checks` in the root `flake.nix` and then add `analysis test` step to `pipeline.yml`.

gromakovsky (commented):
The interface looks ok. Are `.csv` files used anywhere?

analysis/tests/logic_test.py
HellyHenry: To run all tests you run `pytest`. I checked, if I delete `[150, 1000]`, this test fails. Tbh, I plan to write more tests, that's why I added other files, just didn't have time yet and we agreed to have a dummy PR by today.

gromakovsky: Yes, if I run `pytest` I get a nicely looking output and the test fails without that point. So we just need to run `pytest` on CI then.

> Tbh, I plan to write more tests, that's why I added other files, just didn't have time yet and we agreed to have a dummy PR by today.

Ok, got it.

HellyHenry:
Actually, good point with small amount of data, I will definitely fix it

gromakovsky:
@HellyHenry I started integrating it with `edna-server` and realized that I want to change the interface a bit. When an experiment is added, we want to do the following:
1. Call `ic50.py` to compute IC50 parameters for this experiment itself.
2. Call `detect_outliers.py` to compute outliers for the extra sub-experiment.
3. Call `ic50.py` again for this new extra sub-experiment.

It sounds quite inefficient and I believe we can do it in one call. So for each experiment (`(experimentId, [(double, double)])`) I propose to return the following data:
1. `experimentId` as we are doing now.
2. `status` as you added.
3. 4PL parameters as we are doing in `ic50.py`. Note that there we wrap them into `{"Right":}`, but I think now `status` makes more sense and we don't need `Right` then.
4. `outliers` as you added.
5. 4PL parameters for the sub-experiment without outliers. If `outliers` is empty, this field can be omitted.

So I believe some code will need to be shared with `ic50.py` which is a good move anyway. What do you think? Can you do it?

HellyHenry:
@gromakovsky combining everything in one call sounds completely legit. I think I can do it fast enough, but will have to edit ic50.py if you don't mind?

gromakovsky:
So tentative interface (I may update it further while working on the task):
1. Input is a list of objects with the following keys: `experiment` is an integer, `data` is a list of pairs of doubles, `findOutliers` is a boolean flag that specifies whether to find outliers.
2. Output is a list of objects with the following keys: `experiment` is an integer, `status` is a string ("DONE" means that everything is ok, any other string is treated an an error message), `params` is a list of 4PL parameters computed for `data`, `outliers` is a list of indices of outliers (empty if `findOutliers` is `False`), `newParams` is an optional list of 4PL parameters computed for `data` without outliers (not provided iff `outliers` is empty).

> I think I can do it fast enough, but will have to edit ic50.py if you don't mind?

Yes, that's ok, I've already asked Sasha not to touch it in the meantime to avoid conflicts. As a side effect we may get proper error handling in `ic50.py`, it would be great.

HellyHenry:
Nice, thanks for specs! Then I'm getting started

gromakovsky:
> `params` is a list of 4PL parameters computed for `data`

Let's make `params` in the output optional as well. If `status` is not "DONE", there is no `params`.

HellyHenry:
@gromakovsky one concern: with proposed output structure there can be a situation when sigmoid is fitted successfully but outliers can not be calculated. In such case empty outliers list will be returned (so it doesn't break) BUT there will be no warning that something is wrong with calculating outliers. 

Example is when we fit sigmoid on 4 data points. To calculate outliers sigmoids are fitted while omitting one point and it's not possible to fit sigmoid on 3 points. Do we just let it slide (in fact there'll be no factual changes to output) or propose a change? Something like make one more field in output "warning".

gromakovsky:
I think let's just let it slide. This is an extra feature and if we silently don't find any outliers it won't be a problem. Also currently we don't have a neat way to display warnings in UI.

HellyHenry:
Okay!

analysis/README.md
```diff
@@ -0,0 +1,41 @@
...
+The list of features implemented in an analytic module:
+1. Compute parameters of fitted sigmoid.
+2. Detect potential outliers.
+3. Compute parameters of fitted sigmoid without outliers.
+
+### Input format
+JSON containing list of experiments n the form:
+```
+["experiment": int,
+"data": [[float, float], ...]]
```
gromakovsky: Did you intentionally omit `{}` here and in output format?

analysis/pyproject.toml
```diff
@@ -5,10 +5,12 @@ description = ""
 authors = ["Serokell <hi@serokell.io>"]
 
 [tool.poetry.dependencies]
-python = "3.9.*"
+python = ">=3.7.1,<3.10"
 numpy = "^1.20.2"
 scipy = "^1.6.2"
 sklearn = "^0.0"
+pytest = "^6.2.3"
```
gromakovsky: In the second half of [this thread](https://serokell.slack.com/archives/C01F8AQSHGE/p1617113349023000) people told me that test dependencies should go to `dev-dependencies` rather than `dependencies` , so shouldn't `pytest` be moved there?

gromakovsky (commented):
From quick glance looks pretty good. I'm going to test it with my code (#63) and come back here. A couple of extra notes:
1. We should run `pytest` on CI. AFAIU we should expose `test` from `analysis/default.nix`, add `analysis-test = analysis.test` to `checks` in `flake.nix` and add the corresponding step to `pipeline.yml`. If you don't want to delve into this Nix machinery, we can offload it to @mkaito or @balsoft. Also if it takes substantial amount of time I think it's ok to do that in a follow-up PR.
2. We usually write detailed commit messages with `Problem:` and `Solution` clauses and I see you wrote it in the PR description. Please copy it to the commit message and I guess it needs to be updated a bit.

gromakovsky:
Also please delete trailing whitespaces: https://buildkite.com/serokell/edna/builds/583#e82d2c80-0efb-4333-aabd-e6deebeba1b4

analysis/pyproject.toml
HellyHenry: Yeah, seem like a good idea, will move it. Tbh I just haven't checked as PyCharm edits poetry stuff it on its own.

analysis/README.md
HellyHenry: Dunno why I did it like this, will add.

gromakovsky: Shouldn't it be
```
[{"experiment": int,
"data": [[float, float], ...]}]
```
?
maybe even
```
[{"experiment": int,
"data": [[float, float], ...]}, ...]
```

gromakovsky:
> I'm going to test it with my code (#63) and come back here.

As expected, I've found some bugs in my code, investigating/fixing them now.

analysis/README.md
HellyHenry: Oof, I'm kinda retarded, sorry. Sec

HellyHenry:
@gromakovsky I'm completely new to Nix. Probably it's a good idea to ask the guys to move pytest to CI. All the other fixes are pushed.

analysis/README.md
```diff
@@ -1,9 +1,35 @@
...
+1. Compute parameters of fitted sigmoid.
+2. Detect potential outliers.
+3. Compute parameters of fitted sigmoid without outliers.
+### Input format
+JSON containing list of experiments in the form:
+```
+[{"experiment": int,
+"data": [[float, float], ...]},
+...]
+```
```
gromakovsky, line 27: Oh, you might have missed this part:
> `findOutliers` is a boolean flag that specifies whether to find outliers.

So if `"findOutliers": false"` is passed, you shouldn't compute `outliers` and `new_params`. Actually it won't be a problem if you compute them, I just won't read them. But if it's easy to support this flag, let's support it to do less computations.

analysis/README.md
```diff
@@ -1,9 +1,35 @@
...
+"new_params": [float, float, float, float],
+"status": str},
+...]
+```
+`experiment` and `status` are always present in the output. If `"status"` is not `DONE`, other fields are omitted.
+If there are no outliers, `new_params` is omitted. Status may contain error descriptions.
+## Error description
+1. `ERROR: Wrong experiment format.` The experiment from .json provided is in the wrong form.
+2. `ERROR: Too little data.` The least amount of data points to fit sigmoid is 4.
+3. `ERROR: Can not fit sigmoid.` Impossible to fit sigmoid to this particular set of data points.
```
gromakovsky: Why does this file have no empty lines now? :thinking:

HellyHenry: *No one noticed anything...*

analysis/README.md
HellyHenry: That's true, will add it.

analysis/README.md
```diff
@@ -4,6 +4,45 @@ This folder contains implementation of Edna data analysis in Python.
...
+JSON containing list of calculated values per each experiment:
+```
+[{"experiment": int,
+"params": [float, float, float, float],
+"outliers": [int, ...],
+"new_params": [float, float, float, float],
+"status": str},
+...]
+```
+`experiment` and `status` are always present in the output.
```
gromakovsky:
```suggestion
`experiment` and `status` are always present in the output.
```

HellyHenry: why does git not show this particular whitespace beforehand?( Also what's the tool to use to detect them automatically?

balsoft: https://github.com/serokell/serokell.nix/blob/master/overlay/default.nix#L30

gromakovsky: > why does git not show this particular whitespace beforehand?

:shrug: 

> Also what's the tool to use to detect them automatically?

Nearly all IDEs/editors have a builtin option or a plugin to remove trailing whitespaces on save. If you google "remove trailing whitespaces <your editor/IDE>", most likely you'll find a solution.

analysis/ic50_analysis.py
```diff
@@ -0,0 +1,62 @@
...
+            continue
+
+        if find_outliers:
+            try:
+                outliers = detect_outliers(measurements)
+                response_value["outliers"] = outliers
+                if len(outliers) > 0:
+                    new_measurements = np.delete(measurements, outliers, axis=0)
+                    new_params = calculate_sigmoid(new_measurements)
+                    response_value["new_params"] = new_params
```
gromakovsky, line 48: According to my specification, it should be `newParams`. But I can handle `new_params` as well and in fact I've already changed it on my side to make it work. However, now there is inconsistency with `findOutliers`. I think let's use `snake_case` everywhere, so please update `findOutliers` in code and `README.md`.

analysis/README.md
HellyHenry: It was ON, but didn't delete for .md file for some reason(

analysis/README.md
```diff
@@ -4,6 +4,45 @@ This folder contains implementation of Edna data analysis in Python.
...
+1. Compute parameters of fitted sigmoid.
+2. Detect potential outliers.
+3. Compute parameters of fitted sigmoid without outliers.
+
+### Input format
+
+JSON containing list of experiments in the form:
+```
+[{"experiment": int,
+"data": [[float, float], ...],
```
gromakovsky, line 24: One more thing: AFAIU currently the order of points matters, I have an example where I get different results depending on the order. I think we should state that explicitly. So we have 2 options:
1. Say that the order matters and to get good results the points should be sorted. Then I'll add sorting on my side.
2. Say that the order doesn't matter and update your code to sort points.

I am okay with both options, what's your preference?

HellyHenry: Can you please share the example? I haven't seen such behavior yet

HellyHenry: I mean answers can be a bit different but it's a floating point difference. I can sort them explicitly so we always get same results for the matter.

HellyHenry: Anyway, sorted it)

gromakovsky:
```
> python ic50_analysis.py '[{    "experiment": 2,    "data": [[0.00508, 36144], [0.00508, 37086], [0.01524, 36022], [0.01524, 36787], [0.04572, 36248], [0.04572, 36362], [0.13717, 36385], [0.13717, 36724], [0.41152, 35417], [0.41152, 35992], [1.2346, 35461], [1.2346, 35479], [3.7037, 33434], [3.7037, 33917], [11.111, 27498], [11.111, 27737], [33.333, 16689], [33.333, 17041], [100, 5973], [100, 6045], [150, 1000]]}]'

[{"experiment": 2, "status": "DONE", "params": [36482.46722333145, 1.0438910039400358, 41.646644340626764, -7163.367701146651], "outliers": [20], "new_params": [36404.55941244787, 1.1447795315311728, 33.048994525300735, -2541.319176862586]}]
```

```
> python ic50_analysis.py '[{    "experiment": 2,    "data": [[150, 1000], [0.00508, 36144], [0.00508, 37086], [0.01524, 36022], [0.01524, 36787], [0.04572, 36248], [0.04572, 36362], [0.13717, 36385], [0.13717, 36724], [0.41152, 35417], [0.41152, 35992], [1.2346, 35461], [1.2346, 35479], [3.7037, 33434], [3.7037, 33917], [11.111, 27498], [11.111, 27737], [33.333, 16689], [33.333, 17041], [100, 5973], [100, 6045]]}]'             

/home/gromak/Work/biocad/edna/analysis/ic50/__init__.py:10: RuntimeWarning: overflow encountered in power
  return ((a - d) / (1.0 + ((x / c) ** b))) + d
/home/gromak/.cache/pypoetry/virtualenvs/edna-3Rv1oKBx-py3.9/lib/python3.9/site-packages/scipy/optimize/minpack.py:828: OptimizeWarning: Covariance of the parameters could not be estimated
  warnings.warn('Covariance of the parameters could not be estimated',
/home/gromak/Work/biocad/edna/analysis/ic50/__init__.py:10: RuntimeWarning: overflow encountered in power
  return ((a - d) / (1.0 + ((x / c) ** b))) + d
/home/gromak/Work/biocad/edna/analysis/ic50/__init__.py:10: RuntimeWarning: overflow encountered in power
  return ((a - d) / (1.0 + ((x / c) ** b))) + d
/home/gromak/.cache/pypoetry/virtualenvs/edna-3Rv1oKBx-py3.9/lib/python3.9/site-packages/scipy/optimize/minpack.py:828: OptimizeWarning: Covariance of the parameters could not be estimated
  warnings.warn('Covariance of the parameters could not be estimated',
/home/gromak/.cache/pypoetry/virtualenvs/edna-3Rv1oKBx-py3.9/lib/python3.9/site-packages/scipy/optimize/minpack.py:828: OptimizeWarning: Covariance of the parameters could not be estimated
  warnings.warn('Covariance of the parameters could not be estimated',
[{"experiment": 2, "status": "DONE", "params": [36482.46596376745, 1.0438931709031831, 41.6464311832262, -7163.2768968641385], "outliers": [0, 13, 17, 18], "new_params": [36378.55117971187, 1.2019003263313273, 30.035267718626212, -1148.5322663025368]}]
```

gromakovsky:
After sorting there seems to be another bug somewhere.
```
(edna-3Rv1oKBx-py3.9) gromak ~/Work/biotech/edna/analysis ‹gromak/edna84-auto-suspicious↓6➔6|?1⚑4›  % python ic50_analysis.py '[{    "experiment": 2,    "data": [[150, 1000], [0.00508, 36144], [0.00508, 37086], [0.01524, 36022], [0.01524, 36787], [0.04572, 36248], [0.04572, 36362], [0.13717, 36385], [0.13717, 36724], [0.41152, 35417], [0.41152, 35992], [1.2346, 35461], [1.2346, 35479], [3.7037, 33434], [3.7037, 33917], [11.111, 27498], [11.111, 27737], [33.333, 16689], [33.333, 17041], [100, 5973], [100, 6045]]}]'
[{"experiment": 2, "status": "DONE", "params": [36482.46722793712, 1.0438908073333746, 41.646667909718595, -7163.377431449439], "outliers": [0], "new_params": [36404.559361403706, 1.1447795810414698, 33.04899233356807, -2541.3179427101054]}]
(edna-3Rv1oKBx-py3.9) gromak ~/Work/biotech/edna/analysis ‹gromak/edna84-auto-suspicious↓6➔6|?1⚑4›  % python ic50_analysis.py '[{    "experiment": 2,    "data": [[0.00508, 36144], [0.00508, 37086], [0.01524, 36022], [0.01524, 36787], [0.04572, 36248], [0.04572, 36362], [0.13717, 36385], [0.13717, 36724], [0.41152, 35417], [0.41152, 35992], [1.2346, 35461], [1.2346, 35479], [3.7037, 33434], [3.7037, 33917], [11.111, 27498], [11.111, 27737], [33.333, 16689], [33.333, 17041], [100, 5973], [100, 6045], [150, 1000]]}]'
[{"experiment": 2, "status": "DONE", "params": [36482.46722793712, 1.0438908073333746, 41.646667909718595, -7163.377431449439], "outliers": [0], "new_params": [36521.92842614426, 1.0371779543071853, 41.94099610586135, -7336.820829396181]}]
```

In the first case the outlier is the first point, in the second case it's the last one. `params` are equal, but there are 2 problems:
1. In the second case `outliers` should be `[20]`, not `[0]`.
2. `new_params` should be equal in both cases, probably they are different because of the previous point.

HellyHenry:
Definitely my mistake, fixed it and added a respective test.

HellyHenry:
Or not, dfuq is happening. Need several minutes more

analysis/.gitignore
```diff
@@ -0,0 +1,15 @@
...
+venv/
+
+# PyCharm files
+.idea
+idea/
+../.idea/
+
+# Test cache files
+.mypy_cache/
+.pytest_cache/
```
sashasashasasha151: No new line

analysis/ic50/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+        scaled_signal = scaler.fit_transform(measurements[:, 1].reshape(-1, 1)).reshape(-1)
+    else:
+        scaled_signal = scaler.transform(measurements[:, 1].reshape(-1, 1)).reshape(-1)
+
+    starting_point = _get_starting_parameters(measurements[:, 0], scaled_signal)
+    try:
+        with np.errstate(invalid='ignore'):
+            fit_param = curve_fit(_sigmoid, measurements[:, 0], np.array(scaled_signal),
+                                  p0=starting_point)
+    except RuntimeError:
```
sashasashasasha151, line 38: Why don't we catch `ValueError` or `OptimizeWarning`? As I can see they are also possible in `curve_fit`

analysis/ic50/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+        with np.errstate(invalid='ignore'):
+            fit_param = curve_fit(_sigmoid, measurements[:, 0], np.array(scaled_signal),
+                                  p0=starting_point, maxfev=10000)
```
sashasashasasha151, line 41: Does `maxfev` will be the only reason of `RuntimeError`? If it doesn't maybe we can somehow check what error we have and don't run `curve_fit` twice

analysis/ic50_analysis.py
```diff
@@ -0,0 +1,62 @@
...
+        response_value = {"experiment": experiment["experiment"], "status": "DONE"}
+        measurements = np.array(experiment['data'])
+
+        try:
+            find_outliers = experiment["find_outliers"]
+            if type(find_outliers) == str:
+                find_outliers = find_outliers.lower()
+                find_outliers = int(find_outliers == 'true')
+        except KeyError:
+            find_outliers = True
```
sashasashasasha151: Can you please explain it. Why `find_outliers` can be `str`? Why we need this `try/except` here?

analysis/ic50_analysis.py
```diff
@@ -0,0 +1,62 @@
...
+        if np.unique(measurements[:, 0]).shape[0] < 4:
+            response_value["status"] = "ERROR: Too little data."
+            response.append(response_value)
+            continue
+
+        try:
+            params = calculate_sigmoid(measurements)
+            response_value["params"] = params
+        except (ArithmeticError, RuntimeError):
+            response_value["status"] = "ERROR: Can not fit sigmoid."
```
sashasashasasha151, line 37: I think we should somehow show error message. So, my idea was to show full error on backend logs. In this case we want to have full error message in `stderr`. But I don't know if it is good or bad solution. cc @gromakovsky

analysis/ic50/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+    c = concentrations[(np.abs(scaled_measurements - (a + d) / 2)).argmin()]
+    return [a, b, c, d]
+
+
+def _fit_sigmoid(
+        measurements: np.array,
+        scaler: MinMaxScaler, fit_scaler: bool = True,
+) -> Tuple[np.array, np.array, float, MinMaxScaler]:
+    if fit_scaler:
+        scaled_signal = scaler.fit_transform(measurements[:, 1].reshape(-1, 1)).reshape(-1)
```
sashasashasasha151, line 27: How I like this reshapes))) Isn't there any better solution in python?

analysis/ic50/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+            outliers.append(int(sorted_indices[key]))
+    return outliers
+
+
+def calculate_sigmoid(measurements: np.array) -> List[int]:
+    measurements = measurements[np.lexsort((measurements[:, 0], measurements[:, 1]))]
+    scaler = MinMaxScaler()
+    total_fitted_values, sigmoid_params, total_mse, scaler = _fit_sigmoid(measurements, scaler)
+    transformed = scaler.inverse_transform(np.array([sigmoid_params[0], sigmoid_params[3]]).reshape(-1, 1)).reshape(-1)
+    return [transformed[0], sigmoid_params[1], sigmoid_params[2], transformed[1]]
```
sashasashasasha151: No new line

analysis/tests/logic_test.py
```diff
@@ -0,0 +1,96 @@
+import json
```
sashasashasasha151, line 1: Do we need empty `analysis/tests/__init__.py`?

sashasashasasha151: @gromakovsky do we have a task to python tests on CI?

analysis/tests/logic_test.py
```diff
@@ -0,0 +1,96 @@
...
+    except Exception:
+        assert False
+
+
+# noinspection PyBroadException
+def test_sample_data_files(load_sample_data):
+    try:
+        ic50_calc([load_sample_data])
+    except Exception:
+        assert False
```
sashasashasasha151: No new line (and in other files)

sashasashasasha151 (approved):
So, no problems, just some comments, fill free to ignore them or maybe you already discussed them

analysis/ic50_analysis.py
gromakovsky: AFAIU the logic here is to support 3 cases: `find_outliers` is passed as normal `bool`, as a string with `"true` or `"false"` or not passed at all (defaults to `true`). My code should always pass normal bool (`true` or `false`), so `find_outliers = experiment["find_outliers"]` should be enough, but I decided that such generalization is harmless and decided not to comment these lines of code.

> Why we need this `try/except` here?

Actually I have a question regarding `try/except` here, too. Isn't it neater to write `experiment.get("find_outliers", true)`?

analysis/ic50_analysis.py
gromakovsky: I think printing detailed error messages to `stderr` is a good idea. They won't show up in UI, but will show up in backend logs. So `response_value["status"]` is for short error messages that can be shown to user (currently we will show them as is, later we may parse them from Haskell and show differently if we want) and `stderr` is for logs and you can print anything there as long as you don't print _way too much_.

analysis/tests/logic_test.py
gromakovsky: No, but I want to create one.

analysis/.gitignore
```diff
@@ -0,0 +1,15 @@
...
+__pycache__/
+
+# Environments
+.venv
+venv/
+
+# PyCharm files
+.idea
+idea/
+../.idea/
```
gromakovsky: According to https://www.notion.so/serokell/gitignore-3d45b853c4494fa282dc04269f851b2e, files created by your tools should be `.gitignore`'d globally for each person using them instead of `.gitignore`'ing them in each repo.

gromakovsky:
> Tbh, I plan to write more tests, that's why I added other files, just didn't have time yet and we agreed to have a dummy PR by today.

AFAIU `.csv` files are still not used in tests, right? Do you plan to write more tests in this PR and use those files or are they added just for manual testing?

gromakovsky:
One more small request. If I run `python ic50_analysis.py '[{    "experiment": 2,    "data": []}]'` (i. e. pass empty data), I think it's better to return `"status": "ERROR: Too little data."`, but currently it fails with
```
Traceback (most recent call last):
  File "/home/gromak/Work/biotech/edna/analysis/ic50_analysis.py", line 62, in <module>
    main(sys.argv[1:])
  File "/home/gromak/Work/biotech/edna/analysis/ic50_analysis.py", line 25, in main
    if measurements.shape[1] != 2:
IndexError: tuple index out of range
```

analysis/tests/logic_test.py
gromakovsky: https://issues.serokell.io/issue/EDNA-95

analysis/ic50_analysis.py
HellyHenry: Right, I didn't know which format will be passed, so included str just for safety. Regarding `get()` method, it returns `None` if no such key existent. I chose `try/except` because thought, in the future we might want to return an exception if the format of json is incorrect. I don't think there's big of a difference.

analysis/ic50/__init__.py
HellyHenry: That's so common for all ML libraries (in PyTorch you use reshapes all the time for matrices) that I didn't even think of it))))

analysis/tests/logic_test.py
HellyHenry: Empty init file makes python think of a directory as of a package. I think when tests are passed on CI it'll be not relevant, but that's just the standard way for Python.

HellyHenry:
@gromakovsky `.csv` files are tested in `test_sample_data_files`. Maybe that's an overkill, but I wanted to make sure real-world data is processed correctly.

HellyHenry:
I'll add a support for empty 'data' field, sure. Thought it's processed before being passed to Python, but that's not a problem at all.

analysis/ic50_analysis.py
gromakovsky: > Regarding `get()` method, it returns `None` if no such key existent.

[`get()`](https://docs.python.org/3/library/stdtypes.html#dict.get) has another argument that specifies the default value to be returned if no value is present. It defaults to `None`, but I suggest passing `True` as the default value.

gromakovsky:
> .csv files are tested in test_sample_data_files. Maybe that's an overkill, but I wanted to make sure real-world data is processed correctly.

Ah, ok, my bad, I was searching for `Ex3_1.csv` and didn't find it anywhere in code, I could've realized that you don't list all .csv in code but go over them using `os.listdir`. So no problem then.

gromakovsky:
> I'll add a support for empty 'data' field, sure. Thought it's processed before being passed to Python, but that's not a problem at all.

I noticed it while writing tests for py+hs integration that I implemented. There I pass _some data_, I wrote a test with empty data and noticed this behavior :)
I think it makes sense to reject experiments with empty data, but there are multiple places where we should do this check and we may forget it somewhere, so I think it's better to handle it gracefully (I guess it should be straightforward anyway).

analysis/ic50_analysis.py
HellyHenry: So we wouldn't need such exception? Sure, will do it this way

analysis/ic50/__init__.py
HellyHenry: AFAIC yes, the only possible reason for a `RuntimeError` is when it fails to converge in a set amount of iterations. I processed several experiments when it's the case for default 1000 iterations, but it converges on 10k (experimental value, which is balanced between convergence and speed). That's the reason I added a "1 more try" on more iterations (usually it's not invoked, so the speed remains OK). And if it fails the second time, then the exception in main file is triggered.

analysis/ic50/__init__.py
HellyHenry: I added a check for NaN's in data so `ValueError` will not be possible. Regarding `OptimizeWarning` I haven't encountered it, but it's not an exception so doesn't break the code. I can either let it be (so printed as `stderr`) or simply suppress it (as with `np.errstate`) because it carries little information and clogs the output logs.

analysis/ic50_analysis.py
HellyHenry: @sashasashasasha151 Do I simply print it with `print()`? Or go for: `print(*args, file=sys.stderr)`? The second option will just format it as an error. Don't know how backend interprets the output.

analysis/.gitignore
HellyHenry: Configured my global `.gitignore` and deleted this local one, thank you!

analysis/ic50_analysis.py
gromakovsky: Backend reads JSON output from `stdout` and parses it in JSON. If you print something else into `stdout`, it will break parsing. And `stderr` output is catched by the backend and printed into the backend logs. So I think `stderr_print(exc)` is good.

gromakovsky (approved):

gromakovsky:
CI passes in #63. When @sashasashasasha151 approves both PRs, I guess I'll just push #63 branch here and we will merge everything at once.

gromakovsky:
I've added commits from #63 here.

PR #61, opened 2021-04-14
EDNA-77 Add pagination for Library and Dashboard getters

gromakovsky:
## Description

Problem: the number of projects, compounds and other entities stored
in Edna may be quite big sometimes. In this case we don't want to
transfer all entities to frontend at once, it may take quite a lot of
time to load a page.
Also it should be possible to sort entities in tables. We can do it on
frontend if we return all items in one call, but if we return chunks of
data we need to know sorting order.

Solution: add pagination to library and dashboard getters that return lists.
We use `servant-pagination` to achieve that. We get sorting as a bonus.
An alternative is to use `servant-util` that provides sorting and
pagination capabilities. I chose `servant-pagination` because it's on
Hackage, it's mentioned in official Servant docs and this way we can
compare it with `servant-util` that we used before.

Some notes:
1. We still fetch all items from DB and use `applyRange` to filter
them. It's better than what we had because we don't transfer more
data than requested, but it's not the most efficient solution.
I want to start with this solution and we can optimize it as a separate
task.
2. There is a plenty of boilerplate in definitions of `get*s` methods.
I made a number of attempts to avoid it, but didn't succeed. One of the
reasons is that `ExtractRange` and `PutRange` range classes are not
exported so we can't write certain generic functions. Besides, this
generic code becomes extremely complex, so I decided to resort to
boilerplate for now. Perhaps some of this boilerplate should be
encapsulated inside `servant-pagination` itself. If we decide that
it's a good library, we can contribute it there.
3. For experiments we support only `uploadDate` field for now. Ideally we should support sorting by compound, target and methodology as well, but it can't be implemented using simple `applyRange`. So it's left for future work.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-77

- Tests
        silently reappearing again.

No tests because it's done at the API level and we don't have such tests yet. They are being developed in EDNA-57.

- Documentation


backend/src/Edna/Orphans.hs
```diff
@@ -6,21 +6,84 @@ module Edna.Orphans () where
...
+  build = buildForResponse
+
+----------------
+-- Swagger
+----------------
+
+instance ToParamSchema (ContentRange fields resource) where
+  -- TODO SOS HELP NEEDED
+  -- How would you define it? I don't know.
+  toParamSchema Proxy = mempty
```
Martoon-00, line 72: I think this is just a string of specific format?

I see below you use `SwaggerArray`, but from my experience, I think the use case for arrays is different, for instance accepting [multiple query parameters](https://hackage.haskell.org/package/servant-swagger-1.1.10/docs/src/Servant.Swagger.Internal.html#line-300). Single `AcceptRanges` equal to `createdAt, modifiedAt` would also be just a mere `SwaggerString`.

gromakovsky: This is how it looks in Swagger viewer.

image

My problem is that I want to provide some description, but I don't understand what field of `ParamSchema` I need to use to make it appear in `Description` on the screenshot. Probably I'll make another attempt to figure it out when I get some rest from this task :)

Maybe I'll just provide `SwaggerString` type and call it a day because [Content-Range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range) seems to be a standard header anyway.

Martoon-00: Yeah, I remember this problem with swagger :fuuu:

In `servant-util` I sort of resolved this with `ParamDescription` type family that associates a type-level description with the given type (e.g. `ContentRange`) + `SwagerrizeApi` that updates descriptions in most places basing on `ParamDescription`, though this is a fairly big hack.

gromakovsky:
Superseded by #75.

PR #63, opened 2021-04-16
[EDNA-84] Update edna-server to handle new Python analysis

gromakovsky:
## Description

Problem: Python analysis code was updated, now it find suspicious
    points and has a different output format. Our Haskell code doesn't
    handle this new format.
    
Solution:
1. Update `FourPL.hs` to parse new data format and output more data.
2. Add `analyse4PLOne` function that works on one sub-experiment
to simplify usage of analysis in this case.
3. Extract details of Python calling into a separate module in
    preparation for even more Python code.
4. Update Upload service to automatically create an additional
    sub-experiment when at least one suspicious point is found.

Depends on #59.

## Related issue(s)

Resolves https://issues.serokell.io/issue/EDNA-84

- Tests
        silently reappearing again.

- Documentation


gromakovsky:
Alright, I think it's ready now. @sashasashasasha151 you are welcome to review this PR and #59.

backend/src/Edna/Analysis/FourPL.hs
```diff
@@ -70,39 +70,71 @@ instance Buildable AnalysisResult where
...
+-- that we treat as "likely outliers".
+analyse4PL :: [Params4PLReq] -> Edna [(ExperimentId, Either Text Params4PLResp)]
+analyse4PL requests =
+  callPythonAnalysis "ic50_analysis.py" requests >>= mapM respPyToResp
+  where
+    respPyToResp :: Params4PLRespPy -> Edna (ExperimentId, Either Text Params4PLResp)
+    respPyToResp Params4PLRespPy{..} = (plrspyExperiment,) <$> case plrspyStatus of
+      "DONE" -> do
+        plrspParams <- whenNothing plrspyParams $
+          throwM $ PyInvalidFormat "no params in successful computation"
```
sashasashasasha151, line 128: You can use `justOrThrow` but ok

backend/src/Edna/Analysis/Infra.hs
```diff
@@ -0,0 +1,62 @@
...
+  logDebug $ toText $ "Python data request: " <> requestString
+  analysisDir <- liftIO $
+    fromMaybe (".." </> "analysis") <$> lookupEnv "EDNA_ANALYSIS_DIR"
+  (exitCode, out, err) <- liftIO $ readCreateProcessWithExitCode
+    (proc "python3" [pyPath, requestString])
+    { cwd = Just analysisDir, std_out = CreatePipe, std_err = CreatePipe}
+    ""
+  unless (null err) $ logDebug $ "Unexpected python errors: " <> toText err
+  logDebug $ toText $ "Python data response: " <> out
+  unless (exitCode == ExitSuccess) $ throwM $ PyExitFailure exitCode
```
sashasashasasha151: Hm, if not `exitCode == ExitSuccess` then we will see smth in `stderr`, right?

backend/src/Edna/Upload/DB/Query.hs
```diff
@@ -29,36 +32,52 @@ import Edna.Util
...
+            , seResult = val_ (PgJSON result)
+            }
+      in catMaybes
+      [ Just $ se primaryName $ plrspParams <$> eitherResp
+      , (rightToMaybe eitherResp >>= plrspNewSubExp) <&>
+        \(_outliers, params4PL) -> se autoName $ Right params4PL
+      ])
+  where
+    autoName = "Automatic"
+    primaryName = "Primary"
```
sashasashasasha151, line 77: Maybe create ADT for them? (but I am ok with hardcoded values)

sashasashasasha151 (approved):

backend/src/Edna/Analysis/FourPL.hs
gromakovsky: Looks almost the same, so I won't change it here, but will keep it in mind for future.

backend/src/Edna/Analysis/Infra.hs
gromakovsky: Yes, but only if logging mode is debug. Since this case is not normal, I guess include the whole stderr of `python` process into `PyExitFailure` and will print it with debug severity only if `exitCode` is `ExitSuccess` (if it's not success, stderr will be printed by the uncaught exception handler).

backend/src/Edna/Upload/DB/Query.hs
gromakovsky: We still need to convert them to text because users may give arbitrary names to sub-experiments. So I think in most places we'll still operate with `Text` and the benefit of using ADT will be too small.

gromakovsky:
I'll merge this PR along with #59 tomorrow morning.

gromakovsky:
Will be merged together with #59.

PR #66, opened 2021-04-21
[EDNA-85] Refactor py+hs friendship

gromakovsky:
## Description

There is various technical debt after/from #49. This PR consists of relatively small commits fixing each item from that technical debt. 

- [x] `typed-process` instead of `process`
- [x] Pass Python request in `stdin` instead of command line options
- [x] Check whether `python` is properly configured on start
- [x] Add `make run` command
- [x] Improve one test with hard-coded IC50 value

## Related issue(s)

https://issues.serokell.io/issue/EDNA-85

- Tests
        silently reappearing again.

- Documentation


backend/src/Edna/Analysis/Infra.hs
```diff
@@ -50,20 +44,18 @@ instance Exception PythonError where
...
-    logDebug $ "Unexpected python stderr: " <> toText err
-  unless isSuccess $ throwM $ PyExitFailure exitCode err
-  logDebug $ toText $ "Python data response: " <> out
-  either (throwM . PyInvalidFormat . toText) pure $ eitherDecode (encodeUtf8 out)
+  let processConfig =
+        setStdin (byteStringInput requestJSON) $
+        setWorkingDir analysisDir $
+        proc "python3" [pyPath]
+  -- @readProcess_@ automatically captures stdout and stderr.
+  -- It also checks the exit code and throws an exception if it's not 0.
```
sashasashasasha151, line 56: >throws an exception

I don't think that we want to throw exceptions

sashasashasasha151 (approved):

backend/src/Edna/Analysis/Infra.hs
gromakovsky: We were throwing an exception in case of non-zero exit code before this PR as well, the only difference is that we were throwing `PyExitFailure` which doesn't exist anymore. After recent changes in Python code we handle bad cases there explicitly and report them as `"status": "ERROR blabla"`, so `ic50_analysis.py` should exit with 0 for any input data (as long as it can be parsed). If it returns non-zero, it's a bug and I think throwing an exception is the best option in this case (it will be reported as internal server error and it's indeed an internal error, it's a bug in code).

PR #67, opened 2021-04-22
Add reuse lint check to CI and correct licensing information to files

gromakovsky:
## Description

Problem: we want to keep track of copyright and license information in
the repo to make it easier to avoid various legal issues.

Solution: use `reuse lint` to check that all legal information is
present in the repo. Add a step to the CI config to run this command.
Add copyright and license information to all files in the repo.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-15

- Tests
        silently reappearing again.

- Documentation


.gitignore
```diff
@@ -1 +1,5 @@
+# SPDX-FileCopyrightText: 2021 Serokell <https://serokell.io>
+#
+# SPDX-License-Identifier: CC0-1.0
```
worm2fed, line 3: Why the license here is CC0-1.0 and in `backend/.gitignore`, `frontend/.gitignore` it is AGPL-3.0-or-later?

default.nix
```diff
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2021 Serokell <https://serokell.io>
+#
+# SPDX-License-Identifier: CC0-1.0
```
worm2fed, line 3: Why the license here is CC0-1.0 and in `analysis/default.nix`, `backend/default.nix`, `docker.nix`, `flake.nix`, `frontend/default.nix` it is AGPL-3.0-or-later?

shell.nix
```diff
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2021 Serokell <https://serokell.io>
+#
+# SPDX-License-Identifier: CC0-1.0
```
worm2fed, line 3: Why the license here is CC0-1.0 and in `analysis/default.nix`, `backend/default.nix`, `docker.nix`, `flake.nix`, `frontend/default.nix` it is AGPL-3.0-or-later?

worm2fed (changes requested):
I have a few questions about different licenses in similar files. And also there are errors with tsx files. Rest of changes seems ok

gromakovsky:
@worm2fed the general rule is the following: `AGPL-3.0-or-later` applies to all files by default, but some files are likely not copyrightable and the license doesn't matter there, so we set it to `CC0-1.0` to make it more explicit.
https://www.notion.so/serokell/FAQ-8da3477b8b7c4faaab6460ff38fc8929

There is no formal strict rule to distinguish these kinds of files. I put `CC0-1.0` somewhere and maybe I've missed some files. But AFAIU the license doesn't matter in non-copyrightable files anyway, `CC0-1.0` just makes it clearer, but `AGPL-3.0-or-later` shouldn't cause troubles. So even if there are more non-copyrightable files with `AGPL-3.0-or-later``, I suppose it's not a big deal.

All `.gitignore` files now have `CC0-1.0`. Regarding `.nix` files, my logic was the following: there are some files that exist just to make old nix work with flakes, they are small "compatibility shims" copy-pasted everywhere (I mean that each repo that uses flakes with old nix has these files), so they are likely not copyrightable. These files are `default.nix` and `shell.nix` in the root. So they have `CC0-1.0`. Other `.nix` files were written by us from scratch and contain something not entirely trivial, so they have `AGPL-3.0-or-later`.

worm2fed (approved):
Okay, for now it's clear.

kirelagin (approved):
There is a new REUSE recommendation: https://reuse.software/faq/#license-templates.

tl;dr they recommend not to bother with `LicenseRef-` for stupid licenses that embed the copyright notice. Instead they suggest to just include the original unmodified text of the license template (i.e. straight with `<copyright holders>`), since, even if the license says to preserve the copyright notice, the idea is that probably modifying it to be an SPDX header and is acceptable.

PR #68, opened 2021-04-22
[EDNA-92] Docker: Allow backend to print unicode characters

mkaito:
## Description

Backend container would print `?` in place of unicode characters in logs. This was due to a lack of locale archives and not setting a UTF-8 capable locale in the container environment.

I've added glibc locale archives and set `en_US.UTF-8` as active locale. This can be overridden via `-E` flag when creating the container if desired.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-92

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/docker.nix
+++ b/docker.nix
@@ -1,6 +1,8 @@
 { dockerTools
 , backend
 , frontend
+, locale
+, glibcLocales
 , linkFarm
 , runCommand
 , writeTextDir }:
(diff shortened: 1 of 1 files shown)


balsoft (approved):

zhenyavinogradov:
The locales archive is very big, it's 200 MB. There's `C.UTF-8` locale which included in glibc package, which uses UTF-8 encoding as well. What about using this locale instead of `en_US.UTF-8`? Then the locale archive won't have to be included in the image.

PR #70, opened 2021-04-23
Make plots more sleek

sashasashasasha151:
Problem: curve was drawing only by base points, but it looks bad

Solution: generate 1000 points with log scale and compute 4pl for them

## Related issue(s)

https://issues.serokell.io/issue/EDNA-107

- Tests
        silently reappearing again.

- Documentation


gromakovsky (approved):
Plots look better and their rendering is not slow, so I think it's good enough.

PR #71, opened 2021-04-23
[EDNA-57] Add integration tests

worm2fed:
## Description

Problem:
Currently we have some tests for business logic of edna-server, but no tests where endpoints are called. So we can miss some bugs.

Solution:
Add some tests for library which calls our API as client with some mock data and checks response data and status codes for errors.

## Related issue(s)

Resolves https://issues.serokell.io/issue/EDNA-57

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/backend/test/Test/API/LibrarySpec.hs
+++ b/backend/test/Test/API/LibrarySpec.hs
@@ -1,0 +1,292 @@
+-- SPDX-FileCopyrightText: 2021 Serokell <https://serokell.io>
+--
+-- SPDX-License-Identifier: AGPL-3.0-or-later
+
+-- | API testing for Library.
+
+module Test.API.LibrarySpec
+  ( spec
+  ) where
+
+import Universum
+
+import Network.HTTP.Types (badRequest400, notFound404)
+import Servant.Client (ClientEnv)
+import Servant.Client.Core (RunClient)
+import Servant.Client.Generic (AsClientT, genericClient)
+import Servant.Util (fullContent, noSorting)
+import Test.Hspec (Spec, SpecWith, beforeAllWith, describe, it, shouldBe, shouldThrow)
+
+import Edna.Library.Web.API
+  (CompoundEndpoints(..), MethodologyEndpoints(..), ProjectEndpoints(..), TargetEndpoints(..))
+import Edna.Library.Web.Types
+  (CompoundResp(..), MethodologyReq(..), MethodologyResp(..), ProjectReq(..), ProjectResp(..),
+  TargetResp(..))
+import Edna.Util (CompoundId, IdType(..), MethodologyId, ProjectId, SqlId(..), TargetId)
+import Edna.Web.Types (URI, WithId(..))
+
+import Test.API.Util (apiTry, errorWithStatus)
+import Test.Orphans ()
+import Test.SampleData
+import Test.Setup (runWithInit, specWithContextAndEnv)
+
+spec :: Spec
+spec = specWithContextAndEnv $ do
+  beforeAllWith (\(ctx, env) -> runWithInit ctx addSampleData $> env) $ do
+    describe "target API" targetSpec
+    describe "compound API" compoundSpec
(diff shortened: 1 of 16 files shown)


gromakovsky:
I think it looks good, more or less what I had in mind. I would limit this PR to Library to prevent it from becoming too big. And I would rename `Test.API.APISpec` to `Test.API.LibrarySpec`, later we will have to more `Spec` modules in `Test.API`. API tests for _everything_ may take about 1000 lines which is too much for one module. I somewhat dislike duplication between these tests and tests for `Service` modules, but I don't know what's the best solution. I even asked [here](https://serokell.slack.com/archives/CHTNRHX2L/p1619439142009500) :slightly_smiling_face: 

Anyway, I think it's a good start, please wrap it up and I'll rereview. Over time it will hopefully become clearer what to do with this duplication.

worm2fed:
Ok, then it's ready for review.
Yes, we'll receive more clarity in future!

backend/test/Test/API/Util.hs
```diff
@@ -0,0 +1,41 @@
...
+  responseStatusCode, runClientM)
+
+import Edna.Setup (EdnaContext)
+import Edna.Web.API (ednaAPI)
+import Edna.Web.Server (ednaServer)
+
+app :: EdnaContext -> IO Application
+app ctx = return $ serve ednaAPI $ ednaServer ctx
+
+clientEnv :: Int -> IO ClientEnv
```
gromakovsky, line 31: [Here](https://hackage.haskell.org/package/http-client-0.7.8/docs/Network-HTTP-Client.html#v:newManager) they say:
> Creating a new `Manager` is a relatively expensive operation, you are advised to share a single `Manager` between requests instead.

And create it for each individual unit test, it sounds wrong. So I think we should:
1. Add a comment here saying that `clientEnv` is relatively expensive.
2. Update `LibrarySpec` to create it only once. Instead of `SpecWith EdnaContext` we can have `SpecWith (EdnaContext, ClientEnv)`.

gromakovsky: Apparently we can use `aroundAll` to solve this problem.

backend/test/Test/Setup.hs
```diff
@@ -44,19 +50,36 @@ postgresTestServerConnString = lookupEnv postgresTestServerEnvName >>= \case
...
-      connString <- postgresTestServerConnString
-      let testConfig = defaultEdnaConfig &
-            ecDb . dbInitialisation ?~ DbInit EnableWithDrop "./sql/init.sql" &
-            ecDb . dbConnString .~ connString &
-            ecLogging .~ LogNothing
-      runEdna testConfig $ do
-        ctx <- ask
-        liftIO $ callback ctx
+specWithContext :: SpecWith EdnaContext -> Spec
+specWithContext = around withContext
```
gromakovsky: Btw, here we can probably use `aroundAll`, too. We don't need to create `EdnaContext` for each individual test, we can create it once for all tests. But I didn't test it, maybe something will go wrong if we do it.

worm2fed: I'll check it

worm2fed: it's working

backend/test/Test/API/LibrarySpec.hs
```diff
@@ -0,0 +1,292 @@
...
+import Edna.Web.Types (URI, WithId(..))
+
+import Test.API.Util (apiTry, errorWithStatus)
+import Test.Orphans ()
+import Test.SampleData
+import Test.Setup (runWithInit, specWithContextAndEnv)
+
+spec :: Spec
+spec = specWithContextAndEnv $ do
+  beforeAllWith (\(ctx, env) -> runWithInit ctx addSampleData >> pure env) $ do
```
gromakovsky: `a >> pure x` can be written as `a $> x` or `x <$ a` (but I'm also ok with the current version).

gromakovsky (approved):

PR #72, opened 2021-04-26
[EDNA-115] Include primary IC50 into ExperimentResp

gromakovsky:
## Description

Problem: in order to show Dashboard table we query ExperimentsResp
    first and then for each experiment in the response we query its primary
    sub-experiment one-by-one. It's a bit inefficient and we can easily
    decrease number of HTTP endpoint calls.
    
Solution:
1. Add a new field to `ExperimentResp` that holds IC50 computed for the primary sub-experiment. Experiment getter was already querying all sub-experiments, so adding this data to response is pretty straightforward.
2. Update frontend types with this new field and update logic accordingly.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-115

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/frontend/src/api/types.ts
+++ b/frontend/src/api/types.ts
@@ -26,12 +26,12 @@ export type ParsedExcelDto = {
   compounds: ParsedCompoundDto[];
 };
 
-export type ResultDto =
+export type ResultDto<T = number[]> =
   | {
       Left: string;
     }
   | {
-      Right: number[];
+      Right: T;
     };
 
 export type SubExperimentDto = Dto<{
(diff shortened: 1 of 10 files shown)


frontend/src/api/types.ts
```diff
@@ -26,12 +26,12 @@ export type ParsedExcelDto = {
   compounds: ParsedCompoundDto[];
 };
 
-export type ResultDto =
+export type ResultDto<T = number[]> =
```
pva701, line 29: :POG: ARE YOU TYPESCRIPT EXPERT?

pva701 (approved):

frontend/src/api/types.ts
gromakovsky: Just looked at `interface Dto<T, Extra = undefined> {` and did something similar :slightly_smiling_face:

PR #74, opened 2021-04-27
Add backend-supported sorting in Table

pva701:
Problem: we would like to be capable of sorting tables
by single column.

Solution: add machinery for react-table to re-request data for every sorting request.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-46

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -6,57 +6,28 @@ import React, { FunctionComponent, ReactElement, useEffect } from "react";
 import { Route, Switch, useLocation } from "react-router-dom";
 import { Redirect } from "react-router";
 
-import { useRecoilCallback, useRecoilValue, useResetRecoilState, useSetRecoilState } from "recoil";
+import { useRecoilValue, useResetRecoilState } from "recoil";
 import { UploadPage } from "./pages/upload/UploadPage";
 import { LibraryPage } from "./pages/library/LibraryPage";
 import { NotFound } from "./components/NotFound/NotFound";
-import {
-  colorsCounterAtom,
-  compoundIdSelectedAtom,
-  excelFileAtom,
-  experimentsTableSizeAtom,
-  modalDialogAtom,
-  newSubexperimentAtom,
-  projectSelectedIdAtom,
-  selectedSubExperimentsColorAtom,
-  selectedSubExperimentsIdsAtom,
-  targetIdSelectedAtom,
-} from "./store/atoms";
+import { excelFileAtom, modalDialogAtom } from "./store/atoms";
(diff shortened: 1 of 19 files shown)


sashasashasasha151 (changes requested):
image

frontend/src/App.tsx
```diff
@@ -72,9 +75,14 @@ export const App: FunctionComponent = (): ReactElement => {
...
       resetNewSubExperiment();
       resetExperimentsTableSizeAtom();
+      experimentsRefresher();
       // For some weird reason reset doesn't update selectedExperimentsQuery selector
       setSelectedSubExperimentsIds(new Set<number>());
     }
+
+    if (!location.pathname.startsWith("/library")) {
+      libraryRefesher();
+    }
```
sashasashasasha151: So, you have `libraryRefesher` but for `dashboard` you list all refreshers. Why?

frontend/src/components/Table/Table.tsx
```diff
@@ -2,63 +2,112 @@
...
   small,
   collapsible,
+  dataOrQuery,
+  defaultSortedColumn,
 }: LibraryTableProps<T>): React.ReactElement {
-  const { getTableProps, getTableBodyProps, headerGroups, rows, prepareRow } = useTable({
-    columns,
-    data,
-  });
+  const isConstant = isRecoilValue(dataOrQuery);
```
sashasashasasha151: Why `Table` props type is `LibraryTableProps` but not `TableProps`

frontend/src/components/Table/Table.tsx
```diff
@@ -2,63 +2,112 @@
...
+      : dataOrQuery
+  );
+
+  useEffect(() => {
+    if (loadableData.state === "hasValue") {
+      setData(loadableData.contents.concat([]));
+    } else {
+      setData([]);
+    }
+  }, [loadableData]);
```
sashasashasasha151, line 82: Do you need it? Looks likes you use data only in two places and it will be more pretty to just use this check in that places

sashasashasasha151 (changes requested):
image
What is it?

frontend/src/App.tsx
pva701: :putya:

frontend/src/components/Table/Table.tsx
pva701: if you say me how to eliminate cyclic definition in this case (table depends on `data`, `loadableData` depends on table state `sortby`, `data` depends on `loadableData`), I will move this check in that specific space :putya:

sashasashasasha151 (approved):

PR #75, opened 2021-04-27
[EDNA-77] Add pagination+sorting for Library and Dashboard getters

gromakovsky:
## Description

Problem: the number of projects, compounds and other entities stored
in Edna may be quite big sometimes. In this case we don't want to
transfer all entities to frontend at once, it may take quite a lot of
time to load a page.
Also it should be possible to sort entities in tables. We can do it on
frontend if we return all items in one call, but if we return chunks of
data we need to know sorting order.

Solution: add pagination to library and dashboard getters that return lists.
We use `servant-util` and `servant-util-beam-pg` to achieve that.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-77

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/backend/src/Edna/DB/Util.hs
+++ b/backend/src/Edna/DB/Util.hs
@@ -1,0 +1,54 @@
+-- SPDX-FileCopyrightText: 2021 Serokell <https://serokell.io>
+--
+-- SPDX-License-Identifier: AGPL-3.0-or-later
+
+-- | Utility module to write functions that deal with DB.
+
+module Edna.DB.Util
+  ( groupAndPaginate
+  ) where
+
+import Universum
+
+import qualified Data.HashMap.Strict as HM
+import qualified Data.HashSet as HashSet
+import qualified Data.Set as Set
+
+import Lens.Micro (at)
+import Servant.Util (PaginationSpec(..))
+import Servant.Util.Dummy.Pagination (paginate)
+
+import Edna.DB.Schema ()
+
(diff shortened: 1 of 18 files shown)


backend/src/Edna/DB/Util.hs
```diff
@@ -0,0 +1,54 @@
...
+import Lens.Micro (at)
+import Servant.Util (PaginationSpec(..))
+import Servant.Util.Dummy.Pagination (paginate)
+
+import Edna.DB.Schema ()
+
+-- | This function is helpful for queries where we use left join and get
+-- pairs of items. The first item is the main and legendary one, it is requested.
+-- We call it @boka@. Each @boka@ has a list of grandchildren, we call them
+-- @joka@. There can be multiple items with the same @boka@ values. @boka@
```
sashasashasasha151, line 26: :peka:

sashasashasasha151 (approved):
Okay, LGTM, it is really hard in some places to fully understand the logic and find bugs (if they exists)

gromakovsky:
At least I have some tests :harold: I can't say that they are perfect and cover 100% of code, but while I was working on this issue I found a number of bugs due to my tests, so it turned out to be not trivial to pass all tests, which gives us _some confidence_ there are no bugs.

backend/src/Edna/DB/Util.hs
Martoon-00: :kaef: story :joy:

PR #79, opened 2021-04-28
[EDNA-91] Attempt to fix caching issues by disabling cache

mkaito:
## Description

Incorrect caching headers for index.html might result in users seeing an old
version. To fix that, we include a Cache-Control header. Setting it to no-cache
still allows the browser to validate the cache with the upstream server, at the
cost of higher traffic to us.

Here's a link to the Cache-Control docs, for your convenience https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

UPD: we changed this header's value to `no-store`, for details see this PR's discussion.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-91

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/docker.nix
+++ b/docker.nix
@@ -61,6 +61,7 @@ in
         listen 80 default_server;
         root /usr/share/nginx/html;
         location / {
+          add_header Cache-Control 'no-store';
           try_files $uri /index.html =404;
         }
 


zhenyavinogradov (approved):
I tried to look into how nginx caches things, as I understand, with `no-cache` the browser will requery the server to check if the file has changed, and nginx will check ETag the browser provided. But the value of ETag is based on just mtime and size of the file (https://serverfault.com/a/690374), and mtime and nix don't like each other, so there's still a chance to get a false cache hit when the size didn't change. Though I'm not sure how likely is that it will be a problem for us, in any case using `no-cache` should still be an improvement.

gromakovsky:
It doesn't solve the problem for me, unfortunately. I still can reproduce it this way: https://issues.serokell.io/issue/EDNA-91#focus=Comments-96-23664.0-0

@mkaito @zhenyavinogradov can you check whether the same behavior happens for you too? The way I describe in that comment works reliably for me. Here is my branch: https://github.com/serokell/edna/commits/gromak/edna91-investigate

I started with changing copyright year (that appears in the footer) to 2022 and after redeployment it wasn't updated until I clicked `Ctrl+R`. I took Zhenya's comment into account and changed to 20222, but it didn't help. I checked source of the page and noticed that it's quite small and there is almost nothing I could change. But it contained "Edna App" that I changed to "Edna Application". So now when I click "View Page Source" I get really different source ("Edna Application" instead of "Edna App"). But I still get the old page without Ctrl+R :disappointed: :disappointed: :disappointed: :disappointed: :disappointed: :disappointed: :disappointed: :disappointed: :disappointed: :disappointed:

zhenyavinogradov:
It does not happen this way for me with `no-cache`. When I refresh/reopen the page, the browser requeries the url, as expected, and the server either replies `304 Not Modified` or sends the new page, depending on whether the file size changed. I'm not sure why you get this behavior, maybe you need to do Ctrl-Shift-R or clear browser data after `no-cache` is added to nginx config, to make sure cache behavior is reset. I also saw in your screen recording that if you refresh the page to get new content, and then open the same url in another tab, it will show old the old content again, that's really weird too. It could help to see what requests your browser does in network tab in dev tools.

Note that in firefox Ctrl-R is the same as F5 and does not clear the cache, Ctrl-Shift-R does. But firefox also has special behavior when you just focus on the address bar and press enter, or reopen the same url in another tab, in this case it will be more eager to use a cached response (https://superuser.com/a/731024).

Re nginx giving a cached response when file size does not change, it seems to be an actual issue for edna (at least when I test it in my browser), because index.html just links to the main js file and its size almost never changes. We could switch to `Cache-Control: no-store` to fix it. What do you think @mkaito?

mkaito:
I think `no-store` is a bit of a nuclear option. The server would always send the page, regardless of whether it was cached. Then again, this site won't see heavy traffic, so it is an option. I'd much rather figure out why Ivan is seeing what he is seeing though. It does not seem normal behaviour.

There are also some cache busting options in webpack, but I don't remember the details off the top of my head. Basically, filename fingerprinting.

gromakovsky:
Ok, I'm investigating it further. This is what happens when I click Ctrl-Shift-R:
image

The first line is `upload` with type `html`. It looks like html constructed from `index.html` template.

And this is what happens when I just open new tab and open https://staging.edna.serokell.team/upload there:
image

It does not try to load `index.html` from the server, which looks like the first case from https://superuser.com/a/731024

> When pressing enter, the browser may load any resources from cache without rechecking them depending on the expiration time (http `Expires` header) and other http headers. This means that most resources can be loaded as quickly as the disk or RAM can respond, potentially less than 1 ms if the resource is in RAM.

It sounds like we should also take care of other headers, e. g. `Expires`.

gromakovsky:
Meanwhile I don't understand anything. I did this:
image

Then I open a new tab, press Ctrl-Shit-I, open Network tab, then insert https://staging.edna.serokell.team/upload into the address bar and press Enter. And somehow I see this:
image

How is it possible? Where does it take html from if I just cleared all cache?

gromakovsky:
So, before I proceed to further testing I think I need to somehow figure out why Firefox doesn't request html until I click Ctrl-R or F5.

gromakovsky:
Checked in Chromium, it seems to request HTML (gets 304 sometimes).

gromakovsky:
Even if I open private browsing mode in FF, it doesn't load html :confused: 

And this screenshot demonstrates that ETag changes when the size of html changes, but not when js code changes for example:
image


So, AFAIU there are 2 problems:
1. Firefox doesn't make any html request for some reason, even after I clear cache. So `Cache-Control` doesn't make any difference here because there is no request at all. Does it work this way only for me or for someone else as well? Did I clear not all cache? Should I clear cache differently? I'd like to avoid removing all cookies for example.
2. If we make changes that don't change the size of `index.template.html` (that is, 99% of all changes), browser will ask server whether anything has changed and the server will say no.

It's not clear how to solve the first problem, hopefully Zhenya will google something :pled:
For the second problem AFAIU we have 2 options: nuclear one (`no-store`) and something smarter (e. g. we can somehow artificially change the length of html everytime or use cache busting that @mkaito mentioned). Does `no-store` apply only to html content that looks like this?
```
<!doctype html><html lang="en"><head><meta charset="UTF-8"><meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1,user-scalable=0"><title>Edna App</title><link href="/style.28ac312887ad7799a4fd.css" rel="stylesheet"></head><body><div id="root"></div><script src="/app.54de38e607d32c57347f.js"></script></body></html>
```

If that's the case, I think `no-store` is not a problem because this html has just a few hundreds of bytes.

gromakovsky:
image

I don't understand what's going on to be honest. Maybe the problem seems to be even worse.
So I opened https://staging.edna.serokell.team/upload in Chromium from scratch, then I changed the footer (which means that JS code has changed, but the size of index.html and consequently ETag didn't change) and redeployed. Then I opened https://staging.edna.serokell.team/upload again with Network tab in dev tools. `/upload` returns 304 as expected (we don't _want_ it to return 304, but currently it's expected because ETag didn't change) and frontend tries to load old JS script. The new script is `/app.1cafaa765bb2b8658c70.js`, but it loads `/app.44343004e11baafdb9f0.js` (that was in the previous html). And then bad things happen. I checked `http https://staging.edna.serokell.team/app.44343004e11baafdb9f0.js` in my terminal, it returns the same response as you get from root endpoint, just index html. It's not a valid JS and AFAIU that's why I get blank page.

gromakovsky:
So by the moment I finished writing the previous message I think I already understood what's going on. When I get blank page I typically think that something is wrong with my browser/computer/this planet/etc., but apparently the problem is with our code.

Now I'll try to set `no-store`, hopefully it'll resolve the problem.

gromakovsky:
> I checked `http https://staging.edna.serokell.team/app.44343004e11baafdb9f0.js` in my terminal, it returns the same response as you get from root endpoint, just index html.

There's also a problem that we return 200 instead of 404 for non-existing pages, the issue about that is https://issues.serokell.io/issue/EDNA-105.

gromakovsky:
`no-store` seems to solve the second problem. I don't think that the cost of using `no-store` is too high and today is the last day of our work on Edna, so I think we should use `no-store`. I'll update this PR accordingly.

I still don't understand why Firefox doesn't request html and where it takes it from. I don't want to investigate it _now_, I've already spent more time than I planned. I'll wait a bit hoping that @zhenyavinogradov will find something :pled:

zhenyavinogradov:
> Firefox doesn't make any html request for some reason, even after I clear cache.

I believe it should just never happen with `no-cache` (or `no-store`), that is weird. One theory is that it's hard to make firefox clear cache entirely, e.g. maybe to make `no-cache` kick in you also need to restart firefox, or reboot your computer and reinstall windows.

>  I checked http https://staging.edna.serokell.team/app.44343004e11baafdb9f0.js in my terminal, it returns the same response as you get from root endpoint, just index html.

That's how we configured nginx, when it cannot find the file requested, it returns contents of index.html. It makes urls like `/upload` and `/dashboard` work. We could fix it by extracting static files into `/static/` prefix and not falling back to index.html when something under `/static/` is not found.

So the behavior you see in chrome (empty page) is what I would expect with `no-cache`, and that's what I get when I test it myself. It uses index.html with an old .js link, and it tries to interpret an html page as a js script.

Ideally, we would have `no-cache` for index.html, but make nginx validate if the file changed properly, and cache all the other files forever because they have hashes in their names. But `no-store` makes sense if we want to resolve the problems quickly. I'd also wait for @mkaito's opinion.

zhenyavinogradov (approved):

gromakovsky:
> I believe it should just never happen with no-cache (or no-store), that is weird. One theory is that it's hard to make firefox clear cache entirely, e.g. maybe to make no-cache kick in you also need to restart firefox, or reboot your computer and reinstall windows.

Yes, that's a problem, I don't know how to make no-cache kick in and don't want to reinstall windows :stuck_out_tongue: 
So I'll just assume that after we deploy `no-store` and I reinstall windows the problem with firefox will disappear.

> But `no-store` makes sense if we want to resolve the problems quickly.

Yes, that's the case, today is the last day. We can wait for @mkaito's opinion and merge it later (but before the demo), however, we don't have time to implement anything better, so I don't think that @mkaito's opinion will change anything. Merging `no-store` sounds like the best option for now.

PR #80, opened 2021-04-29
[EDNA-76] DB concurrency fixes

worm2fed:
## Description

Problem:
1. When uploadFile is called, we add various data to DB using multiple INSERT statements. We put them into transact. My idea was to achieve atomicity by this transact, but apparently it doesn't work the way I want. For example: add threadDelay $ 30 * 1000 * 1000 after insertTarget and try to upload a file. Once you click "Add experiments", go to the library. You'll see targets there, but no compounds (yet). AFAIU it means that other threads can observe the state when a file is only partially uploaded. For example, it's possible that a sub-experiment is already added, but removed measurements are not.
2. Extension of the previous one: try killing edna-server with SIGINT during threadDelay $ 30 * 1000 * 1000. If you run it again, you'll see that targets have been added and can be seen, but the rest is not. It's a more serious problem than (1) because (1) normally may last only for ≈a second, but this problem is persistent.

The root of problem:
We have three functions: `withConnection`, `transact`  and `runPg`. Function `transact` supposed to wrap some code inside transactions, and it’s working almost as expected. If we look at postgresql log we can see that transactions starts with process id `X` and ends with same one, but actions inside transaction has another id `Y`.
So, the problem is that when we call transact we use withConnection and particular actions uses withConnection too, which produces different processes.

Solution:
Store functions to work with connections in `EdnaContext` instead of `ConnPoll`, and then when we need to execute some actions in transaction - override that functions to use single connection.

## Related issue(s)

Resolves https://issues.serokell.io/issue/EDNA-76

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/backend/src/Edna/DB/Connection.hs
+++ b/backend/src/Edna/DB/Connection.hs
@@ -3,19 +3,23 @@
 -- SPDX-License-Identifier: AGPL-3.0-or-later
 
 module Edna.DB.Connection
-  ( ConnPool (..)
-  , withPostgresConn
+  ( PostgresConn(..)
   , createConnPool
   , destroyConnPool
+  , postgresConnPooled
+  , postgresConnSingle
+  , withPostgresConn
   ) where
 
 import Universum
 
 import Control.Concurrent (threadDelay)
 import Control.Exception.Safe (catchIOError)
-import Data.Pool (Pool, createPool, destroyAllResources)
+import Data.Pool (Pool, createPool, destroyAllResources, withResource)
 import Data.Time.Clock (nominalDay)
 import Database.Beam.Postgres (Connection, close, connectPostgreSQL)
+import Database.PostgreSQL.Simple.Transaction (withSavepoint, withTransactionSerializable)
+import RIO (MonadUnliftIO, withRunInIO)
 
 import Edna.Config.Definition (EdnaConfig, dbConnString, dbMaxConnections, ecDb)
 import Edna.Util (ConnString(..), logUnconditionally)
@@ -63,3 +67,42 @@ createConnPool (ConnString connStr) maxConnsNum = liftIO $ ConnPool <$>
 -- | Destroys a @ConnPool@.
 destroyConnPool :: MonadIO m => ConnPool -> m ()
 destroyConnPool (ConnPool pool) = liftIO $ destroyAllResources pool
+
+-- | Provides methods for working with connections to Postgres database.
+data PostgresConn = PostgresConn
+  { pcWithConnection  :: forall (m :: Type -> Type) a. MonadUnliftIO m
+                      => (Connection -> m a) -> m a
+  , pcWithTransaction :: forall (m :: Type -> Type) a. MonadUnliftIO m
+                      => Connection -> m a  -> m a
+  }
+
+-- | Implementation of 'PostgresConn' which uses a 'ConnPool' to support
+-- multithreading. Methods have the following properties:
+--
+-- * 'pcWithConnection' blocks if no idle connections are available and the
+--   maximum number of DB connections has been reached.
+-- * 'pcWithTransaction' uses @withTransactionSerializable@ as implementation.
+postgresConnPooled :: ConnPool -> PostgresConn
+postgresConnPooled (ConnPool pool) = PostgresConn
+  { pcWithConnection  = \action ->
+      withRunInIO $ \unlift ->
+        withResource pool $ unlift . action
+  , pcWithTransaction = \conn action ->
+      withRunInIO $ \unlift ->
+        withTransactionSerializable conn $ unlift action
+  }
+
+-- | Implementation of 'PostgresConn' which simply uses the single connection.
+-- It's only intended to be constructed inside @transact@ in order to make sure
+-- that all database actions wrapped with @transact@ are performed with a
+-- single connection.
+--
+-- and 'pcWithTransaction' implementation is @withSavepoint@ to support
+-- "nested" transactions
+postgresConnSingle :: Connection -> PostgresConn
+postgresConnSingle conn = PostgresConn
+  { pcWithConnection  = ($ conn)
+  , pcWithTransaction = \conn' action ->
+      withRunInIO $ \unlift ->
+        withSavepoint conn' $ unlift action
+  }
--- a/backend/src/Edna/DB/Integration.hs
+++ b/backend/src/Edna/DB/Integration.hs
@@ -21,7 +21,6 @@ import Universum
 
 import qualified Database.Beam.Postgres.Conduit as C
 
-import Data.Pool (withResource)
 import Database.Beam.Backend.SQL.BeamExtensions (runInsertReturningList)
 import Database.Beam.Backend.SQL.Row (FromBackendRow)
 import Database.Beam.Postgres (Connection, Pg, Postgres, runBeamPostgres, runBeamPostgresDebug)
@@ -30,21 +29,24 @@ import Database.Beam.Query
   (SqlDelete, SqlInsert, SqlSelect, SqlUpdate, runDelete, runInsert, runSelectReturningList,
   runSelectReturningOne, runUpdate)
 import Database.Beam.Schema (Beamable)
-import Database.PostgreSQL.Simple.Transaction (withTransactionSerializable)
-import RIO (withRunInIO)
 
-import Edna.DB.Connection (ConnPool(..))
+import Edna.DB.Connection (PostgresConn(..), postgresConnSingle)
 import Edna.Logging (logUnconditionally)
-import Edna.Setup (Edna, edConnectionPool, edDebugDB)
+import Edna.Setup (Edna, EdnaContext(_edDBConnection), edDBConnection, edDebugDB)
 
 withConnection :: (Connection -> Edna a) -> Edna a
 withConnection action = do
-  ConnPool pool <- view edConnectionPool
-  withRunInIO $ \unlift -> withResource pool (unlift . action)
+  c <- view edDBConnection
+  pcWithConnection c action
 
 transact :: Edna a -> Edna a
-transact action = withConnection $
-  \conn -> withRunInIO $ \unlift -> withTransactionSerializable conn (unlift action)
+transact action = do
+  c <- view edDBConnection
+  pcWithConnection c $ \conn ->
+    pcWithTransaction c conn $ do
+      let c' = postgresConnSingle conn
+      local (\ctx -> ctx { _edDBConnection = c' }) $ do
+        pcWithTransaction c' conn action
 
 runPg :: Pg a -> Edna a
 runPg pg = withConnection $ \conn ->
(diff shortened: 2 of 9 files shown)


worm2fed:
@gromakovsky I've fixed the main problem.

In task description we additionally have following:
> There are some functions (e. g. updateProject) where we read some data from DB and then update DB depending on this data. The problem is that some action may happen between read and update and cause uncaught SQL error (which will be shown as Internal Server Error which is not ok) or maybe even invalid state (if we don't check something by SQL constraint). For example, in case of updateProject 2 people may concurrently change name of 2 different projects to X. One of these calls may fail with SqlError in this case. One possible solution is to do UPDATE in one query with subquery that checks that no projects with such name exist.

So, I think it's ok for now. We have mechanism to catch SqlError when there name duplications, and then we throw err400. 
Of course we can wrap everything in transactions, then if one user starts updating project and another one wants to update project too - he shall wait until first one commits transactions or rolls it back.
It described more detailed here: https://wiki.postgresql.org/wiki/SSI

backend/src/Edna/DB/Connection.hs
```diff
@@ -63,3 +67,42 @@ createConnPool (ConnString connStr) maxConnsNum = liftIO $ ConnPool <$>
 -- | Destroys a @ConnPool@.
 destroyConnPool :: MonadIO m => ConnPool -> m ()
 destroyConnPool (ConnPool pool) = liftIO $ destroyAllResources pool
+
+-- | Provides methods for working with connections to Postgres database.
+data PostgresConn = PostgresConn
+  { _withConnection  :: forall (m :: Type -> Type) a. MonadUnliftIO m
```
gromakovsky: According to our general convention I think it's better to call it `_pcWithConnection`. Also is `_` necessary? I think it's usually used for somewhat private things or when there is another thing with the same name without `_`. But here these things don't look private in any sense and apparently don't conflict with anything (especially if we add `pc` prefix).

UPD: I see that `_withConnection` conflicts with `withConnection` and in this sense it's somewhat private indeed. So maybe `_` is justified, but I still think we can just replace it with `pc` prefix.

backend/src/Edna/DB/Connection.hs
```diff
@@ -63,3 +67,42 @@ createConnPool (ConnString connStr) maxConnsNum = liftIO $ ConnPool <$>
...
+                     => (Connection -> m a) -> m a
+  , _withTransaction :: forall (m :: Type -> Type) a. MonadUnliftIO m
+                     => Connection -> m a  -> m a
+  }
+
+-- | Implementation of @PostgresConn@ which uses a @ConnPool@ to support
```
gromakovsky: Here I think you can use single quotes `'` instead of `@` because `PostgresConn` and `ConnPool` are actual identifiers in code. And below you should write `@withConnection@` instead of ``` `withConnection` ``` because these are not real identifiers (or write `'_withConnection'` for example).

https://github.com/ghc/haddock/blob/master/doc/cheatsheet/haddocks.pdf

backend/src/Edna/DB/Integration.hs
```diff
@@ -30,21 +29,24 @@ import Database.Beam.Query
...
+  _withConnection c action
 
 transact :: Edna a -> Edna a
-transact action = withConnection $
-  \conn -> withRunInIO $ \unlift -> withTransactionSerializable conn (unlift action)
+transact action = do
+  c <- view edDBConnection
+  _withConnection c $ \conn ->
+    _withTransaction c conn $ do
+      local (\ctx -> ctx { _edDBConnection = postgresConnSingle conn }) $ do
```
gromakovsky: Might be simpler to write `let c' = postgresConnSingle conn`, then `ctx { _edDBConnection = c' }` here and just use `c'` below (without `view`). But that's up to you.

gromakovsky (changes requested):
The fix looks good, I have only cosmetic comments. I'm going to test it now.

gromakovsky:
> I'm going to test it now.

Alright, it seems to work correctly :clap:

gromakovsky:
I think it'd be nice to also add a test. We can cause an error inside `uploadFile'` somehow and check that nothing is added to DB in this case. For example, we can pass `error` as `fileName` to it, expect exception in the test and then check that no compounds and targets are added. If you do it this way, please also add a comment to `uploadFile'` saying that there is a test which relies on the fact that we add experiment file after compounds and targets.

backend/src/Edna/DB/Connection.hs
worm2fed: I'll fix it

gromakovsky:
> We have mechanism to catch SqlError when there name duplications, and then we throw err400.

AFAIU currently we don't explicitly catch `SqlError` anywhere, so it will be an internal server error in this case. We can catch `SqlError`, but I slightly dislike it because:
1. If we catch any `SqlError` and just return 400 in this case, we may catch an unexpected error that happens for a reason other than bad request. It's hard to predict what errors may happen when you make SQL query. For example, we can get an error if we change the DB schema and forget to migrate/wipe existing DB, in this case we shouldn't respond with 400. So ideally we should distinguish `SqlError`s.
2. However, if we want to distinguish `SqlError`s, we need to parse textual value attached to it and it doesn't sound very robust. I'm afraid that these textual values may change and they may even be different for different versions of PostgreSQL. It's unlikely, but still sounds possible.

Another problem is that some constraint may be not enforced at the level of SQL: maybe we forgot to add such a constraint or it's just too hard to express in terms of SQL. Then if we check it only in Haskell and then insert some data after the check, we may insert incorrect data if there was a change between the check and the insertion.

So I think ideally we should update functions like `updateProject` (which do `SELECT` and `UPDATE` or `INSERT`) to contain one transaction: either using `transact` or by grouping `SELECT` and `UPDATE` into one query (where it's possible).

worm2fed:
@gromakovsky I've got it.

> So I think ideally we should update functions like updateProject (which do SELECT and UPDATE or INSERT) to contain one transaction: either using transact or by grouping SELECT and UPDATE into one query (where it's possible).

What about `getProject`? In this case I think we should wrap it in `transact` too, because we one user can start updating and another one can request this project

gromakovsky:
> What about `getProject`? In this case I think we should wrap it in `transact` too, because we one user can start updating and another one can request this project

`getProject` looks ok, it makes just one query, there is no point in adding `transact` to it.

Oh, maybe you mean `addProject`, then yes, it makes sense there.

gromakovsky:
> I think it'd be nice to also add a test. 

Please do it and I'll approve.

worm2fed:
@gromakovsky done

backend/test/Test/UploadSpec.hs
```diff
@@ -101,6 +107,19 @@ spec = withContext $ startWithInitial $ do
...
+
+    it "rollbacks targets and compounds due processing error" $ \ctx -> do
+      runRIO ctx (uploadFile' (SqlId 1) (SqlId 1) "" (error "e") "" sampleFile)
+        `shouldThrow` anyErrorCall
+
+      targets <- runRIO ctx $ Library.getTargets Nothing Nothing Nothing
+      compounds <- runRIO ctx $ Library.getCompounds Nothing Nothing Nothing
+      experiments <- runRIO ctx $ Dashboard.getExperiments Nothing Nothing Nothing
+
+      -- We do not use 0 here, because we add some data in upper tests
```
gromakovsky: Actually it's trickier, you shouldn't rely on which tests will run before. For example if you pass `--match rollback` to the test-suite, this test will fail because it will run only this test. When I was writing such tests I tried to carefully structure them, so that no matter what pattern we pass to `--match`, all tests will succeed. Specifically, it's better to put tests that don't mutate DB before mutating tests. In this case it should solve the problem, just move this test before `"successfully adds targets and compounds"` and test that all numbers below are 0.

If you do it, you can also define ```shouldBeZero l = liftIO $ length l `shouldBe` 0``` and use `runRIO ctx` once instead of 3 times:
```
runRIO ctx $ do
  Library.getTargets Nothing Nothing Nothing >>= shouldBeZero
  Library.getCompounds Nothing Nothing Nothing >>= shouldBeZero
  Dashboard.getExperiments Nothing Nothing Nothing >>= shouldBeZero
```

backend/test/Test/UploadSpec.hs
```diff
@@ -101,6 +107,19 @@ spec = withContext $ startWithInitial $ do
...
+      length targets `shouldBe` 4
+      length compounds `shouldBe` 5
+      length (wItem <$> erExperiments experiments) `shouldBe` 10
```
gromakovsky: Btw, `wItem <$>` here looks weird. I mean, length of `erExperiments experiments` must be equal to the length of `(wItem <$> erExperiments experiments)`, `<$>` applied to a list can't change its length.

backend/test/Test/UploadSpec.hs
```diff
@@ -101,6 +107,19 @@ spec = withContext $ startWithInitial $ do
...
+      runRIO ctx (uploadFile' (SqlId 1) (SqlId 1) "" (error "e") "" sampleFile)
+        `shouldThrow` anyErrorCall
+
+      targets <- runRIO ctx $ Library.getTargets Nothing Nothing Nothing
+      compounds <- runRIO ctx $ Library.getCompounds Nothing Nothing Nothing
+      experiments <- runRIO ctx $ Dashboard.getExperiments Nothing Nothing Nothing
+
+      -- We do not use 0 here, because we add some data in upper tests
+      length targets `shouldBe` 4
+      length compounds `shouldBe` 5
```
gromakovsky: Also currently these checks don't test the actual problem, because even if you add `sampleFile` successfully, you won't add new compounds/targets because they are already added.

gromakovsky:
I reverted your "Fix transactions execution" commit and all tests still pass, can you please check why?

backend/test/Test/UploadSpec.hs
worm2fed: That was first I've did. But unfortunately, after transaction rollback postgres increases auto_increment index.
I can put it in separate describe section with `startWithInitial`

backend/test/Test/UploadSpec.hs
worm2fed: agree

worm2fed:
I'll check it

backend/test/Test/UploadSpec.hs
worm2fed: executing `_ <- ctx <$ runWithInit ctx addInitialData` before other actions inside tests works as well

gromakovsky: Oh, indeed, there is a problem. Okay, `runWithInit ctx addInitialData` looks good to me.

backend/test/Test/UploadSpec.hs
```diff
@@ -109,17 +109,15 @@ spec = withContext $ startWithInitial $ do
         onlyDisabled secondaryMeasurements `shouldBe` secondaryDisabled
 
     it "rollbacks targets and compounds due processing error" $ \ctx -> do
+      _ <- ctx <$ runWithInit ctx addInitialData
```
gromakovsky: `ctx <$ ` causes this expression to return `ctx` and you ignore it. So `ctx <$` is redundant.

worm2fed:
> I reverted your "Fix transactions execution" commit and all tests still pass, can you please check why?

the problem was that tests was running with 'dirty' database, after fixes with tests it fail (in case of revert).
image

gromakovsky (approved):
LGTM

PR #81, opened 2021-04-29
Compute mean IC50

pva701:
## Related issue(s)

https://issues.serokell.io/issue/EDNA-119

- Tests
        silently reappearing again.

- Documentation


frontend/src/store/selectors.ts
```diff
@@ -175,25 +176,39 @@ export const selectedSubExperimentsQuery = selector<SubExperimentWithMeasurement
...
-        "Left" in acc
-          ? acc
-          : "Left" in x.meta.item.result
-          ? { Left: x.meta.item.result.Left }
-          : { Right: acc.Right + x.meta.item.result.Right[2] },
-      { Right: 0 }
-    );
-    if ("Left" in ic50Sum) return ic50Sum;
-    return { Right: ic50Sum.Right / subs.length };
+    return evalMeanIC50(get(selectedSubExperimentsMetaQuery).map(x => x.item.result));
```
gromakovsky: The issue says:
> Somewhere in Dashboard we show mean IC50 for all primary sub-experiments of all experiments where at least one sub-experiment is selected. I think that's what they want, not mean for all selected sub-experiments.

Currently we calculate average for all sub-experiments shown on plot, but I think they want each selected experiment to have equal weight, even if multiple sub-experiments are selected. They say:
> Необходим расчет среднего значения IC50 при выборе нескольких экспериментов

It's formulated imprecisely, so maybe I am wrong. Maybe it's better to have the following logic: if multiple sub-experiments are chosen from one experiment, we calculate average for them and then we take average of all these "averages". So if you have `se1` and `se2` from one experiment and `se3` for another experiment, we can return `((se1 + se2) / 2 + se3) / 2`. I don't insist on doing that and I don't know what they want. If you think I am saying nonsense, feel free to ignore. In general please try to maximize usefulness of this value divided by complexity of implementing it.

pva701: >So if you have se1 and se2 from one experiment and se3 for another experiment, we can return ((se1 + se2) / 2 + se3) / 2

Ok, maybe it makes sense... it's not hard to implement

frontend/src/components/IC50Line/IC50Tooltip.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { Loadable } from "recoil";
+import React from "react";
+import { ResultDto } from "../../api/types";
+import { Tooltip } from "../Tooltip/Tooltip";
+import { formatIC50 } from "../../utils/utils";
+import "./IC50Tooltip.scss";
+
+export type IC50Value = ResultDto<number | number[]> | Loadable<ResultDto<number | number[]>>;
+
+interface IC50LineProps {
```
sashasashasasha151: Tooltip props

frontend/src/components/IC50Line/IC50Tooltip.tsx
```diff
@@ -0,0 +1,67 @@
...
+  ic50: IC50Value;
+  className?: string;
+}
+
+export function IC50Tooltip({ ic50, className }: IC50LineProps): React.ReactElement {
```
sashasashasasha151: You write Tooltip here 4 times, but they are same all time. Like you can just return from some function all information which you need for Tooltip and write only one Tooltip

sashasashasasha151 (approved):

frontend/src/components/IC50Line/IC50Tooltip.tsx
pva701: nichego ne ponyal

gromakovsky (approved):

PR #83, opened 2021-05-01
[EDNA-89] Support sorting of experiments by more fields

gromakovsky:
## Description

Problem: it's possible to sort experiments by upload date, but there
are other columns in the table with experiments and it often makes
sense to sort by them as well.

Solution:
- [x] Support sorting by compound, target and methodology names.
- [x] Add these names to `ExperimentResp`.
- [x] Add implicit base sorting by IDs, so that items with equal sorting keys (or all items when there is no sorting) always have deterministic order.
- [x] Update frontend according to new (extended) `ExperimentResp` structure.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-89

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/backend/src/Edna/Dashboard/DB/Query.hs
+++ b/backend/src/Edna/Dashboard/DB/Query.hs
@@ -25,8 +25,8 @@ import Database.Beam.Backend (SqlSerial(..))
 import Database.Beam.Postgres (PgJSON(..), Postgres)
 import Database.Beam.Postgres.Full (deleteReturning)
 import Database.Beam.Query
-  (Q, QExpr, aggregate_, all_, as_, cast_, countAll_, guard_, int, leftJoin_, lookup_, select,
-  subquery_, update, val_, (&&.), (<-.), (==.))
+  (Q, QExpr, aggregate_, all_, as_, cast_, countAll_, guard_, int, just_, leftJoin_, lookup_,
+  select, subquery_, update, val_, (&&.), (<-.), (==.))
 import Fmt (pretty)
 import Servant.Util (HList(..), PaginationSpec, (.*.))
 import Servant.Util.Beam.Postgres (sortBy_)
@@ -36,10 +36,11 @@ import Edna.Analysis.FourPL (AnalysisResult, Params4PL(..))
 import Edna.DB.Integration
   (runDeleteReturningList', runSelectReturningList', runSelectReturningOne', runUpdate')
 import Edna.DB.Schema (EdnaSchema(..), ednaSchema)
-import Edna.DB.Util (groupAndPaginate)
+import Edna.DB.Util (groupAndPaginate, sortingSpecWithId)
 import Edna.Dashboard.DB.Schema
 import Edna.Dashboard.Web.Types (ExperimentResp(..), ExperimentSortingSpec)
 import Edna.ExperimentReader.Types (FileMetadata)
+import Edna.Library.DB.Schema (TestMethodologyRec, TestMethodologyT(..))
 import Edna.Setup (Edna)
 import Edna.Upload.DB.Schema (ExperimentFileT(..))
 import Edna.Util as U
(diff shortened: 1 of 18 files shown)


frontend/src/api/types.ts
```diff
@@ -94,9 +94,9 @@ export type CompoundDto = {
 
 export type ExperimentDto = Dto<{
   project: number;
-  compound: number;
-  target: number;
-  methodology?: number;
+  compound: [number, string];
```
pva701, line 107: why do you need `id` here?

gromakovsky: I am not sure if we need it, but I decided to keep it here because it's very small and may be useful in future. This way the response is more self-contained.

backend/src/Edna/Dashboard/DB/Query.hs
```diff
@@ -135,19 +164,29 @@ getExperiments mProj mComp mTarget sorting pagination =
         )
       )
   where
-    sortingApp ((_experimentRec, _projectId, _methodologyId, uploadDate, _subExpId), _) =
+    eExperimentId' :: ExperimentRec -> SqlSerial Word32
+    eExperimentId' = eExperimentId
```
worm2fed: Why do we need it?

backend/src/Edna/Orphans.hs
```diff
@@ -28,3 +29,9 @@ instance ApiHasArgClass (MultipartForm' mods tag t) where
   apiArgName _ = "multipart"
 
 instance ApiCanLogArg (MultipartForm' mods tag t)
+
+-- defined the same way as in @Lens.Micro.Internal@ (where instances are
+-- provided only for tuples with up to 5 items)
+instance Field1 (a,b,c,d,e,f,g) (a',b,c,d,e,f,g) a a' where
```
worm2fed:
```suggestion
instance Field1 (a, b, c, d, e, f, g) (a', b, c, d, e, f, g) a a' where
```

backend/src/Edna/Orphans.hs
```diff
@@ -28,3 +29,9 @@ instance ApiHasArgClass (MultipartForm' mods tag t) where
...
+  _1 k ~(a,b,c,d,e,f,g) = (,b,c,d,e,f,g) <$> k a
```
worm2fed:
```suggestion
  _1 k ~(a, b, c, d, e, f, g) = (, b, c, d, e, f, g) <$> k a
```

backend/src/Edna/Dashboard/DB/Query.hs
gromakovsky: I was getting a weird compilation error involving complex Beam types without it :shrug: 

Let me check whether it's still necessary.

gromakovsky: No longer necessary, removed.

frontend/src/utils/utils.ts
```diff
@@ -49,8 +49,49 @@ export function zip<A, B>(a: A[], b: B[]): [A, B][] {
...
+// We show at most 5 characters. We empirically determined that 5 characters
+// can be fit into the IC50 column.
+// If we return a string that doesn't fit into the column for some reason,
+// we have fallback logic in CSS that limits the width and adds ellipsis.
 export function formatIC50(x: number): string {
-  return x.toFixed(3);
+  // We are carefully listing all cases here to think about each of them.
+  // Perhaps it can be optimized, e. g. using 'toPrecision', but I (@gromak)
+  // decided to follow the KISS principle.
+  if (x >= 1e5) {
```
worm2fed: Maybe it's better to use switch or lookup map?
https://betterprogramming.pub/stop-putting-so-many-if-statements-in-your-javascript-3b65aaa4b86b

worm2fed (changes requested):

frontend/src/utils/utils.ts
gromakovsky: How to use `switch` here? AFAIU switch works when you compare with some values for equality, but here I compare only for `>=`. Same for lookup map. I am new to JS/TS, so maybe I'm missing something, but it's not clear to me how to use `switch` here.

worm2fed: switch:
```
values=[1000,  2000 ... 29000, 30000];

switch(true) {
  case (val < values[0]): /* do something */ break;
  case (val < values[1]): /* do something */ break;
  ...
  case (val < values[29]): /* do something */ break;
}
```

with lookup map.. maybe there are option to use expressions as keys.. but maybe I'm wrong and this is from other language

gromakovsky: Hm, ok, I'll give it a try.

worm2fed (approved):

frontend/src/utils/utils.ts
worm2fed: it's good now (and faster)

PR #85, opened 2021-05-03
[EDNA-95] Run pytest on CI

gromakovsky:
## Description

Problem: we have tests for Python code, but CI doesn't run them,
so they may silently break.

Solution: add a new CI step that calls `pytest` in `nix-shell`.

## Related issue(s)

Resolve https://issues.serokell.io/issue/EDNA-95

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -11,6 +11,11 @@ steps:
     commands:
       - nix-shell --run 'reuse lint'
 
+# Checks for the analysis
+  - label: analysis test
+    commands:
+      - nix-shell --run 'cd analysis && pytest'
+
 # Checks for the backend
   - label: validate cabal files
     commands:


gromakovsky:
@mkaito I tried the simplest possible approach that came to my mind and it seems to work. Is it good enough?

Alternatively, I can add some `test` to `analysis/default.nix` and then have `checks.analysis-test` in `flake.nix`, but I don't know how to define this `test`. So if you think we should have `test` in `analysis/default.nix`, please help me to define it. Currently we just export env built by `mkPoetryEnv`. AFAIU I should return an object with `env = mkPoetryEnv …` and `test = ???`

mkaito:
Our verification steps should all be encoded as flake checks, as per Elagin. I'd have to look into how Poetry works on Nix, I'm not familiar (yet).

kirelagin:
That’s exactly how it works, yes, you get the stuff from the Poetry env on your PATH and then run `pytest`. Just make it into a check instead of running `nix-shell` in the pipeline config directly.

PR #87, opened 2021-05-03
[EDNA-97] Switch to openapi3

worm2fed:
Problem:
We use swagger2 library that follows Swagger specification version 2. For quite a while a newer specification is available and called OpenAPI 3. AFAIK it's sufficiently mature and all reasonable tools can handle it already. We should use it because it's more modern.
Besides, in the openapi3 library ToParamSchema class returns Schema while in swagger2 it returns ParamSchema. ParamSchema has less fields, in particular it seems to have no Description field, so we can't provide a description for response headers for example.

Solution:
Update resolver to 17.10 because of dependencies pain (for example `servant-swagger-ui` was refactored recently to use more general Value type, which allows us to use openapi3 with swagger ui, and updated package included to newer snapshot).
The rest of changes is just updating declarations according to new standard and defining some instances.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-97

- Tests
        silently reappearing again.

- Documentation


backend/test/Test/SwaggerSpec.hs
```diff
@@ -8,7 +8,7 @@ module Test.SwaggerSpec
 
 import Universum
 
-import Servant.Swagger (validateEveryToJSON)
+import Servant.OpenApö (validateEveryToJSON)
```
balsoft: Ärë ÿöü sürë?

backend/src/Edna/Analysis/FourPL.hs
```diff
@@ -66,8 +66,7 @@ instance FromJSON Params4PL where
       constr (a, b, c, d) = Params4PL a b c d
 
 instance ToSchema Params4PL where
-  declareNamedSchema Proxy =
-    declareNamedSchema @Params4PLTuple Proxy
+  declareNamedSchema Proxy = declareNamedSchema @Params4PLTuple Proxy
```
maksbotan, line 69: BTW, why do you derive this schema from a tuple and not from record (using generic mechanism)? Is it more convenient for front to operate on tuples?

I suppose the question is for @gromakovsky

backend/src/Edna/Upload/Web/Types.hs
```diff
@@ -51,16 +51,29 @@ instance Buildable FileBS where
...
-        & schema .~ ParamOther (mempty
-            & in_ .~ ParamFormData
-            & paramSchema .~ (mempty & type_ ?~ S.SwaggerFile))
+      body :: O.RequestBody
+      body = mempty
+        & O.content .~ content
+        & O.required ?~ True
+
+      content :: Map.InsOrdHashMap MediaType O.MediaTypeObject
+      content = Map.singleton "multipart/form-data" $ mempty
```
maksbotan: Instead of `Map.singleton` you can do

```haskell
content = mempty
  & at "multipart/form-data" ?~ mempty & schema ?~ paramSchema
```

backend/src/Edna/Upload/Web/Types.hs
```diff
@@ -51,16 +51,29 @@ instance Buildable FileBS where
...
+        & O.schema ?~ paramSchema
+
+      paramSchema :: O.Referenced O.Schema
+      paramSchema = O.Inline $ mempty
+        & O.type_ ?~ O.OpenApiObject
+        & O.required .~ ["file"]
+        & O.properties .~ props
+
+      props :: Map.InsOrdHashMap Text (O.Referenced O.Schema)
+      props = Map.singleton "file" $ O.Inline $ mempty
```
maksbotan: likewise, I'd use `at` and merge props directly into `paramSchema`, w/o creating additional binding, like

```haskell
  & O.properties .~ (mempty
    at "file" ?~ (mempty & ....)
```

backend/src/Edna/Util.hs
```diff
@@ -170,21 +168,15 @@ schemaOptions :: SchemaOptions
...
-    :: ( Generic a
-       , S.GToSchema (G.Rep a)
-       , GenericHasSimpleShape a "genericDeclareNamedSchemaUnrestricted" (GenericShape (G.Rep a))
-       )
-    => Proxy a -> Declare (S.Definitions S.Schema) S.NamedSchema
-gDeclareNamedSchema = S.genericDeclareNamedSchema schemaOptions
+gDeclareNamedSchema :: (Generic a, O.GToSchema (G.Rep a))
+                    => Proxy a
+                    -> Declare (O.Definitions O.Schema) O.NamedSchema
+gDeclareNamedSchema = O.genericDeclareNamedSchema schemaOptions
```
maksbotan, line 174: btw, it's a lot easier to use `DerivingVia` for that: https://github.com/biocad/web-template/blob/master/src/Web/Template/Servant/Aeson.hs#L14

I don't suggest doing anything like that now, that would require changing a lot of code, I'm only demonstrating a concept.
cc @gromakovsky

backend/src/Edna/Upload/Web/Types.hs
worm2fed: I was looking for something like this! thanks

backend/src/Edna/Upload/Web/Types.hs
worm2fed: hm...
@maksbotan now I have a lot of errors:

```log
src/Edna/Upload/Web/Types.hs:63:26: error:
    • Could not deduce (Lens.Micro.Internal.At
                          (Map.InsOrdHashMap MediaType O.MediaTypeObject))
        arising from a use of ‘at’
      from the context: HasOpenApi api
        bound by the instance declaration
        at src/Edna/Upload/Web/Types.hs:54:10-71
    • In the first argument of ‘(?~)’, namely
        ‘at "multipart/form-data"’
      In the second argument of ‘(&)’, namely
        ‘at "multipart/form-data" ?~ mempty’
      In the first argument of ‘(&)’, namely
        ‘mempty & at "multipart/form-data" ?~ mempty’
   |
63 |       content = mempty & at "multipart/form-data" ?~ mempty
   |                          ^^^^^^^^^^^^^^^^^^^^^^^^

src/Edna/Upload/Web/Types.hs:63:29: error:
    • Could not deduce (IsString
                          (Lens.Micro.Internal.Index
                             (Map.InsOrdHashMap MediaType O.MediaTypeObject)))
        arising from the literal ‘"multipart/form-data"’
      from the context: HasOpenApi api
        bound by the instance declaration
        at src/Edna/Upload/Web/Types.hs:54:10-71
    • In the first argument of ‘at’, namely ‘"multipart/form-data"’
      In the first argument of ‘(?~)’, namely ‘at "multipart/form-data"’
      In the second argument of ‘(&)’, namely
        ‘at "multipart/form-data" ?~ mempty’
   |
63 |       content = mempty & at "multipart/form-data" ?~ mempty
   |                             ^^^^^^^^^^^^^^^^^^^^^

src/Edna/Upload/Web/Types.hs:63:54: error:
    • Could not deduce (Monoid
                          (Lens.Micro.Internal.IxValue
                             (Map.InsOrdHashMap MediaType O.MediaTypeObject)))
        arising from a use of ‘mempty’
      from the context: HasOpenApi api
        bound by the instance declaration
        at src/Edna/Upload/Web/Types.hs:54:10-71
    • In the second argument of ‘(?~)’, namely ‘mempty’
      In the second argument of ‘(&)’, namely
        ‘at "multipart/form-data" ?~ mempty’
      In the first argument of ‘(&)’, namely
        ‘mempty & at "multipart/form-data" ?~ mempty’
   |
63 |       content = mempty & at "multipart/form-data" ?~ mempty
   |                                                      ^^^^^^

src/Edna/Upload/Web/Types.hs:65:11: error:
    • Could not deduce (O.HasSchema
                          (Map.InsOrdHashMap MediaType O.MediaTypeObject)
                          (Maybe (O.Referenced O.Schema)))
        arising from a use of ‘O.schema’
      from the context: HasOpenApi api
        bound by the instance declaration
        at src/Edna/Upload/Web/Types.hs:54:10-71
    • In the first argument of ‘(?~)’, namely ‘O.schema’
      In the second argument of ‘(&)’, namely ‘O.schema ?~ paramSchema’
      In the expression:
        mempty & at "multipart/form-data" ?~ mempty
          & O.schema ?~ paramSchema
   |
65 |         & O.schema ?~ paramSchema
   |           ^^^^^^^^
```

maksbotan: That's what you get for using obscure packages like `microlens` :)
The required instances are present in `lens`. Can you clarify with @gromakovsky whether it's ok to use `Control.Lens.at` instead of microlens variant?

backend/src/Edna/Analysis/FourPL.hs
gromakovsky: Hm, yes, I think I assumed that it would be easier for frontend to handle tuples, but I don't remember why I assumed that and maybe it was wrong assumption. Or maybe I decided to use tuples to have smaller JSON :shrug: 

@worm2fed if you want, you can check how this value is handled by frontend and maybe update it if you want.

backend/src/Edna/Upload/Web/Types.hs
```diff
@@ -14,19 +14,19 @@ module Edna.Upload.Web.Types
 
 import Universum
 
-import qualified Data.Swagger as S
+import qualified Data.HashMap.Strict.InsOrd as Map
```
gromakovsky: Usually `Map` is used for `Data.Map` or `Data.Map.Strict`. If I see `Map.*` in code, I usually assume that it's from `containers`. So I would expand this name, maybe `InsMap` for example.

backend/src/Edna/Upload/Web/Types.hs
gromakovsky: > The required instances are present in `lens`. 

I think they are present somewhere else because `lens` doesn't seem to depend on `insert-ordered-containers`. But I don't want to search for these instances.

> whether it's ok to use `Control.Lens.at` instead of `microlens` variant?

Yes, I think it's ok to use anything from `lens` because we transitively depend on `lens` anyway.

backend/src/Edna/Web/Swagger.hs
```diff
@@ -52,13 +52,9 @@ type SwaggerUI =
...
 -- | Attach an UI serving given documentation to the given server.
-withSwaggerUI
-  :: Proxy api
-  -> S.Swagger
-  -> Server api
-  -> Server (WithSwaggerUI api)
-withSwaggerUI _ swagger server =
-  server :<|> swaggerSchemaUIServer swagger
+withSwaggerUI :: Proxy api -> O.OpenApi -> Server api
+              -> Server (WithSwaggerUI api)
```
gromakovsky: I don't know if we have any formal strict rules in style guide about indentation in this case, but you can see that in all examples in style guide we use fixed indentation. Example:
image

As you can see, there are always 2 spaces no matter what the length of `putValueInState` is. The advantage of this approach is that if we rename `putValueInState` we won't need to modify lines of code below `putValueInState`.

In case of our code (`withSwaggerUI`) this indentation forces us to increase/decrease indentation on this line if we rename `withSwaggerUI`.

backend/src/Edna/Upload/Web/Types.hs
maksbotan: I meant they are defined for `Control.Lens.At`, not present in `lens`, sorry. Instances come from `insert-ordered-containers` and are transitively visible here from `openapi3` itself.

backend/src/Edna/Analysis/FourPL.hs
worm2fed: ok

backend/src/Edna/Web/Swagger.hs
worm2fed: I've got. The reason I've used such indention, that if I put `:: ...` on newline, vscode stops to highlight function name ( but this is issue to plugin
I'll update it

backend/src/Edna/Dashboard/DB/Query.hs
```diff
@@ -164,6 +163,9 @@ getExperiments mProj mComp mTarget sorting pagination =
         )
       )
   where
+    eExperimentId' :: ExperimentRec -> SqlSerial Word32
+    eExperimentId' = eExperimentId
```
gromakovsky: Can you please double check that it's actually needed in the final version of code?

gromakovsky (approved):
Don't forget to cleanup the commit history.

backend/src/Edna/Dashboard/DB/Query.hs
worm2fed: well, now it compiles without it))

PR #88, opened 2021-05-04
[EDNA-128] Add edna-generator

gromakovsky:
## Description

Problem: for testing we often need to add some sample data to DB,
doing it manually is tedious and for stress testing it's almost
infeasible.

Solution: add a new edna-generator executable capable of generating
some hardcoded data and saving it to DB.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-128

- Tests
        silently reappearing again.

No tests because this utility is just for developers. We may use it in a benchmark at some point.

- Documentation

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -26,6 +26,8 @@ ifeq (${EDNA_USE_CABAL},1)
 		--color $(TEST_ARGUMENTS)
 	call_server = cabal new-run $(PACKAGE):edna-server $(CABAL_BUILD_OPTIONS) -- \
 		$(RUN_ARGUMENTS)
+	call_generator = cabal new-run $(PACKAGE):edna-generator $(CABAL_BUILD_OPTIONS) -- \
+		$(RUN_ARGUMENTS)
 	call_clean = cabal new-clean
 else
 	call_build = stack build $(STACK_BUILD_OPTIONS) --file-watch $(PACKAGE)
@@ -33,6 +35,8 @@ else
 	call_test = stack build $(STACK_TEST_OPTIONS) $(PACKAGE) \
 		--test-arguments "--color $(TEST_ARGUMENTS)"
 	call_server = stack run edna-server -- \
+		$(RUN_ARGUMENTS)
+	call_generator = stack run edna-generator -- \
 		$(RUN_ARGUMENTS)
 	call_clean = stack clean $(PACKAGE)
 endif
@@ -54,6 +58,12 @@ test:
 run:
 	cd ../analysis && poetry install && \
 		poetry run bash -c "cd ../backend && $(call call_server)"
+
+# Run edna-generator inside `poetry shell` making sure Python analysis will work.
+# If you want to pass additional arguments, use `RUN_ARGUMENTS` env variable.
+run-generator:
+	cd ../analysis && poetry install && \
+		poetry run bash -c "cd ../backend && $(call call_generator)"
 
 clean:
 	$(call call_clean)
--- a/backend/README.md
+++ b/backend/README.md
@@ -48,3 +48,19 @@ NOTE: priority of configurations is following: command line arguments, environme
 To run tests you need to have [`pg_tmp`](http://eradman.com/ephemeralpg/) to run DB for tests.
 Then run `make test` to execute tests.
 This command automatically executes [`poetry`](https://python-poetry.org/) to obtain Python dependencies.
+
+## Generate sample data
+
+Apart from `edna-server` executable we provide `edna-generator`.
+This tool can generate sample data for Edna and put it into DB.
+You can build and run it the same way as `edna-server` (via `stack` or `cabal`).
+You can also run it using [`Makefile`](./Makefile).
+It has the same options as `edna-server` plus some additional options that specify how much data to generate.
+Pass `--help` to get the full list of available options.
+
+You are adviced to set logging to `prod` or `nothing` (otherwise there will be too much output in logs)
+Also maybe you also want to run it on empty DB.
+Example with `prod` logging and empty DB: `RUN_ARGUMENTS='-c dev-config.yaml -l prod --init-mode enable-with-drop' make run-generator`.
+If DB is not empty and there are conflicts (such as already existing project name), by default they will be skipped.
+
+Note that `edna-generator` is an experimental tool for developers and may be unstable/unreliable sometimes.
--- a/backend/generator/Generator.hs
+++ b/backend/generator/Generator.hs
@@ -1,0 +1,162 @@
+-- SPDX-FileCopyrightText: 2021 Serokell <https://serokell.io>
+--
+-- SPDX-License-Identifier: AGPL-3.0-or-later
+
+module Generator
+  ( GeneratorSpec (..)
+  , generatorDetails
+  , generateAndSave
+  ) where
+
+import Universum
+
+import qualified Data.ByteString.Lazy as BSL
+import qualified Data.Map.Strict as Map
+import qualified Data.Text as Text
+
+import Control.Exception.Safe (throwString)
+import Options.Applicative.Help.Pretty (Doc, indent, linebreak)
+
+import qualified Edna.Library.Service as Library
+import qualified Edna.Upload.Service as Upload
+
+import Edna.ExperimentReader.Types
+  (FileContents(..), FileMetadata(..), Measurement(..), TargetMeasurements(..))
+import Edna.Init (initEdna)
+import Edna.Library.Error (LibraryError(..))
+import Edna.Library.Web.Types (MethodologyReq(..), ProjectReq(..))
+import Edna.Logging (logMessage)
+import Edna.Setup (Edna)
+import Edna.Util (SqlId(..))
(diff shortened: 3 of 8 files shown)


backend/generator/Generator.hs
```diff
@@ -0,0 +1,110 @@
...
+        in (compoundName, measurements gsMeasurementsNum projectIdx methodologyIdx)
+
+measurements :: Word -> Word32 -> Word32 -> [Measurement]
+measurements n project methodology = map measurement [1 .. n]
+  where
+    measurement i =
+      let concentration = 200 / 2 ^ i in Measurement
+      { mConcentration = concentration
+      , mSignal =
+        realToFrac project / (1 + (concentration / realToFrac methodology)) +
```
worm2fed: I think it'll be more readable if you add `signal` to let block, similar to `concentration`

backend/generator/Generator.hs
```diff
@@ -0,0 +1,110 @@
...
+      }
+
+addExperiments :: GeneratorSpec -> Edna ()
+addExperiments spec@GeneratorSpec {..} = do
+  forM_ [1 .. gsProjectNum] $ \projectIdx -> do
+    forM_ [1 .. gsMethodologyNum] $ \methodologyIdx ->
+      addExperiment projectIdx methodologyIdx
+    logMessage $ "Processed project " <> show projectIdx
+  where
+    addExperiment :: Word32 -> Word32 -> Edna ()
```
worm2fed: Maybe we should create type for `Word32`? for example `type Idx = Word32`. I see that we use same value as number of elements and as indexes.. but we use it as num only in `GeneratorSpec` and there we can leave definition as it is and use `Idx` only in other function's type signatures

backend/generator/Main.hs
```diff
@@ -0,0 +1,73 @@
...
+import Edna.Config.Prepare (prepareConfig)
+import Edna.Setup (runEdna)
+
+import Generator (GeneratorSpec(..), generateAndSave)
+
+main :: IO ()
+main = withUtf8 $ do
+  (options, generatorSpec) <- execParser parserInfo
+  config <- prepareConfig options
+  -- let generatorSpec = GeneratorSpec 5 8 6 7 12
```
worm2fed: do you need it?

backend/Makefile
```diff
@@ -26,6 +26,8 @@ ifeq (${EDNA_USE_CABAL},1)
 		--color $(TEST_ARGUMENTS)
 	call_server = cabal new-run $(PACKAGE):edna-server $(CABAL_BUILD_OPTIONS) -- \
 		$(RUN_ARGUMENTS)
+	call_generator = cabal new-run $(PACKAGE):edna-generator $(CABAL_BUILD_OPTIONS) -- \
+		$(RUN_ARGUMENTS)
```
worm2fed, line 30: what will happen with `edna-server` if we pass `--projects 4`? As I can see below we call both `edna-server` and `edna-generator` with same arguments (`RUN_ARGUMENTS`), but generator arguments is unknown for `edna-server`

backend/README.md
```diff
@@ -48,3 +48,13 @@ NOTE: priority of configurations is following: command line arguments, environme
...
 Then run `make test` to execute tests.
 This command automatically executes [`poetry`](https://python-poetry.org/) to obtain Python dependencies.
+
+## Generate sample data
+
+Apart from `edna-server` executable we provide `edna-generator`.
+This tool can generate sample data for Edna and put it into DB.
+You can build and run it the same way as `edna-server` (via `stack` or `cabal`).
+You can also run it using [`Makefile`](./Makefile).
+It has the same options as `edna-server` plus some additional options that specify how much data to generate.
```
worm2fed, line 58: Can you add comment please, that to see all available options we can call `edna-generator --help`?

worm2fed (changes requested):
There are few style moments and one concern about arguments in Makefile. The rest seems ok

backend/Makefile
gromakovsky: `--projects` is parsed only by `edna-generator`, not by `edna-server`. We can have `RUN_GENERATOR_ARGUMENTS` or just `GENERATOR_ARGUMENTS` for `edna-generator`, but I don't see a problem in using `RUN_ARGUMENTS` for both. This way there are less variable names to remember. You run only one of these things, so there should be no ambiguity. Semantic of `RUN_ARGUMENTS` is that they are passed to the executable we want to run, no matter what executable it is.

backend/generator/Generator.hs
gromakovsky: Okay, I'll define `type Idx = Word32` and use in this module. If you suggest defining it in the library, we may consider doing that, but it's a broader topic and out of scope for this PR.

backend/README.md
gromakovsky: Done.

backend/generator/Main.hs
gromakovsky: No, forgot about it. Removed.

backend/generator/Generator.hs
gromakovsky: Done.

backend/generator/Generator.hs
gromakovsky: Done.

backend/generator/Generator.hs
worm2fed: > If you suggest defining it in the library, we may consider doing that, but it's a broader topic and out of scope for this PR.

I saw that we have some complex types in tests, and maybe in other places too. I think we should pay attention to it and define some types, we can create an issue to research it and make according changes

backend/Makefile
worm2fed: I just tried to call this: `edna-server --projects` and there are an error:
```
Invalid option `--projects'

Usage: edna-server [-c|--config EDNA_CONFIG]
...
```

worm2fed: I can try to clone this branch and see what happens if I pass some arguments

worm2fed: oh, I can see it now, we do not use generator in tests. Dunno why I thought we use it)

gromakovsky: > I just tried to call this: `edna-server --projects` and there are an error:

Yes, that's right, as I said, there is no `--projects` option in `edna-server`.
`RUN_ARGUMENTS="--projects 5" make run-generator` makes sense, while `RUN_ARGUMENTS="--projects 5" make run` doesn't. When you type `RUN_ARGUMENTS` you know what command you want to run, so you can write appropriate arguments. So I don't understand the problem. Why would you do `RUN_ARGUMENTS="--projects 5" make run`? You can treat `make run` and `make run-generator` as 2 functions both having an argument call `RUN_ARGUMENTS`. Ideally you would write `make run --run-arguments {arguments}` and `make run-generator --run-arguments {arguments}`, but AFAIU it can't be done easily, so we use env vars instead.

Though if your workflow is to set `RUN_ARGUMENTS` once in your `.bashrc` (or something equvalent) and never touch it, then your suggestion makes sense to me, then you can set `RUN_ARGUMENTS` and `RUN_GENERATOR_ARGUMENTS` to different values and never touch them. Though in this case we should probably add `EDNA` prefix to these variables if you set them globally for you, not just locally for `make`.

gromakovsky: So what's the use case when it's a problem for you?

worm2fed: well, for now it's not a problem.. but if we'll want to use generator along with tests or run some 'stress test' with `make test` it can be a problem..

what do you think?

worm2fed: I guess we can resolve this conversation, because:
> oh, I can see it now, we do not use generator in tests. Dunno why I thought we use it)

but we have to come back to this question if
> we'll want to use generator along with tests or run some 'stress test' with `make test`

worm2fed (approved):

worm2fed:
@gromakovsky can we pass `--project 0` and `--targets 10`?

gromakovsky:
Yes, but it's somewhat useless. If you pass `--projects 0`, no experiments will be added because it's impossible to add experiments without projects. So only methodologies will be added in this case (if their number is greater than 0).

worm2fed (commented):
Well, it's ok now

gromakovsky:
The bad case is if you pass `--targets 0`, because in this case there will be ~overflow~ underflow in `[0 .. gsTargetNum - 1]`. I didn't bother writing checks for `GeneratorSpec` because this tool is only for developers know are supposed to know what they are doing, but since you asked and it's easy to add this check, I'll add it (for `gsCompoundNum` and `gsTargetNum`).
`gsProjectNum = 0` case is useful if you want to add only methodologies for some reason, so I won't prohibit it.

gromakovsky:
> we'll want to use generator along with tests or run some 'stress test' with make test

I think for stress tests usually benchmarks are used, so we may have a benchmark that will use `Generator` module. Then we'll probably have `make bench` command. I'll create an issue about benchmarks and stress testing after this PR is merged.

As I wrote in the PR description:
> No tests because this utility is just for developers. We may use it in a benchmark at some point.

PR #91, opened 2021-05-06
[EDNA-117] Generate proper MDe links

worm2fed:
Problem:
Currently we are generating stub MDe links. In practice they should look like https://HOST/list/27 where 27 is the number of compound.
While normally compounds are called CN<number (e. g. CN27), some compounds have more complex names (like CN27-ab.cde:1). But its not clean how to generate links for them.

Solution:
Store optional MDe link for each compound on backend and make it possible to manually edit it, just like we do it for ChemSoft links.
If no link is provided, we should do the following:
- for CN<number> use <number> as part of link
- for compound names in different format we'll just show that the link is not known.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-117

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/backend/src/Edna/Config/Definition.hs
+++ b/backend/src/Edna/Config/Definition.hs
@@ -14,6 +14,7 @@ module Edna.Config.Definition
   , ecApi
   , ecDb
   , ecLogging
+  , ecMdeHost
 
   , ApiConfig (..)
   , acListenAddr
@@ -30,6 +31,8 @@ module Edna.Config.Definition
 
   , LoggingConfig (..)
   , parseLoggingConfig
+
+  , MdeHost
   ) where
 
 import Universum
@@ -69,6 +72,8 @@ data LoggingConfig =
   -- ^ No logging.
   deriving stock (Generic, Show, Eq)
 
+type MdeHost = Text
+
 -- | Parse LoggingConfig
 parseLoggingConfig :: e -> String -> Either e LoggingConfig
 parseLoggingConfig err p =
@@ -101,6 +106,7 @@ data EdnaConfig = EdnaConfig
   { _ecApi :: ApiConfig
   , _ecDb :: DbConfig
   , _ecLogging :: LoggingConfig
+  , _ecMdeHost :: Maybe MdeHost
   } deriving stock (Generic, Show)
 
 defaultEdnaConfig :: EdnaConfig
@@ -115,6 +121,7 @@ defaultEdnaConfig = EdnaConfig
     , _dbInitialisation = Nothing
     }
   , _ecLogging = LogProd
+  , _ecMdeHost = Nothing
   }
 
 ---------------------------------------------------------------------------
--- a/backend/src/Edna/Config/Environment.hs
+++ b/backend/src/Edna/Config/Environment.hs
@@ -11,28 +11,33 @@ module Edna.Config.Environment
   , dbInitialisationModeEnv
   , dbMaxConnectionsEnv
   , loggingEnv
+  , mdeHostEnv
   ) where
 
 import Universum
 
-import Control.Exception (throwIO)
 import qualified Data.Char as C
-import System.Environment (lookupEnv)
+import qualified Data.Text as T
 import qualified Text.Show
 
-import Edna.Config.Definition (LoggingConfig(..), parseLoggingConfig)
+import Control.Exception (throwIO)
+import System.Environment (lookupEnv)
+
+import Edna.Config.Definition (LoggingConfig(..), MdeHost, parseLoggingConfig)
 import Edna.Util (ConnString(..), DatabaseInitOption, NetworkAddress, parseDatabaseInitOption)
 
 data EnvParseError = WrongApiListenAddr
                    | WrongDbMaxConnections
                    | WrongDbInitialisationMode
                    | WrongLogging
+                   | WrongMdeHost
 
 instance Show EnvParseError where
   show WrongApiListenAddr = "Can't parse EDNA_API_LISTEN_ADDR"
   show WrongDbMaxConnections = "Can't parse EDNA_DB_MAX_CONNECTIONS"
   show WrongDbInitialisationMode = "Can't parse EDNA_DB_INITIALISATION_MODE"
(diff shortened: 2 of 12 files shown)


worm2fed:
I'll add tests tomorrow

backend/src/Edna/Upload/Service.hs
```diff
@@ -130,11 +133,23 @@ uploadFile' projSqlId@(SqlId proj) methodSqlId@(SqlId method)
       measurementsToSummary fileMeasurements
 
 insertTarget :: Text -> Edna (Text, TargetId)
-insertTarget targetName = (targetName,) . fromSqlSerial . tTargetId <$> LQ.insertTarget targetName
+insertTarget targetName = do
+  (targetName,) . fromSqlSerial . tTargetId <$> LQ.insertTarget targetName
+
+
```
gromakovsky: Normally there is 1 empty line between functions, 2 empty lines should be used only in special cases and I think this one is not special.

backend/src/Edna/Config/Definition.hs
```diff
@@ -69,6 +72,8 @@ data LoggingConfig =
   -- ^ No logging.
   deriving stock (Generic, Show, Eq)
 
+type MdeHost = String
```
gromakovsky: By default we prefer `Text` to `String` and use `String` only where it's required by external API or there is a solid reason to use it.

backend/src/Edna/Upload/Service.hs
```diff
@@ -130,11 +133,23 @@ uploadFile' projSqlId@(SqlId proj) methodSqlId@(SqlId method)
...
+buildMdeLink :: Maybe MdeHost -> Text -> Maybe Text
+buildMdeLink Nothing _ = Nothing
```
gromakovsky: If below `host` is `Nothing`, you will get `Nothing` in result. Did you write this case as an optimization to avoid checking whether the string ends with a number in case `host` is `Nothing`? If yes, then it seems to make sense.

Also instead of pattern matching you can do `<- host` inside `Maybe` monad. Something like:
```

buildMdeLink mHost compoundName = do
  host <- mHost
  let (_, ns) = T.break isNumber compoundName
  guard (not $ T.null ns && T.all isNumber ns)
  Just . (<> "/" <> ns) . T.pack $ host
```

This way you will check whether `host` is `Just` only once.

gromakovsky (commented):
Looks good, waiting for tests. And please test it manually as well if you didn't already.

backend/src/Edna/Config/Definition.hs
worm2fed: I thought about it, but in task there:
> Add optional configuration parameter mde-link, its type is string
I'll change it to Text)

backend/src/Edna/Upload/Service.hs
worm2fed: >  Did you write this case as an optimization to avoid checking whether the string ends with a number in case host is Nothing?

yes

> This way you will check whether host is Just only once.

you are right, I'll change it

backend/src/Edna/Config/Definition.hs
gromakovsky: Oh, I meant abstract string type, not specifically Haskell's `String` :slightly_smiling_face: So it's string from the user's standpoint as the user doesn't care what actual type used in Haskell, users just pass it as command line option/env var/config field.

backend/src/Edna/Config/Environment.hs
```diff
@@ -110,3 +115,6 @@ dbInitialisationInitScriptEnv = lookupEnv "EDNA_DB_INITIALISATION_INIT_SCRIPT"
 
 loggingEnv :: IO (Maybe LoggingConfig)
 loggingEnv = parseEnv parseLoggingConfig "EDNA_LOGGING" WrongLogging
+
+mdeHostEnv :: IO (Maybe MdeHost)
+mdeHostEnv = lookupEnv "EDNA_MDE_HOST" <&> (T.pack <$>)
```
gromakovsky: In `Universum` there is `<<$>>` which is `fmap . fmap`, so I guess it can be written as `T.pack <<$>> lookupEnv "EDNA_MDE_HOST"`.

backend/test/Test/UploadSpec.hs
```diff
@@ -121,6 +126,25 @@ spec = withContext $ startWithInitial $ do
         Library.getCompounds noSorting fullContent >>= shouldBeZeroLength
         Dashboard.getExperiments Nothing Nothing Nothing noSorting fullContent
           >>= shouldBeZeroLength . erExperiments
+
+    -- Note that there are some checks in LibrarySpec and DashboardSpec as well.
```
gromakovsky: Erm, did you copy this comment intentionally? I mean, there are some tests with file upload where compounds are added, but we don't check anything about mde link generation there :shrug:

backend/test/Test/UploadSpec.hs
```diff
@@ -121,6 +126,25 @@ spec = withContext $ startWithInitial $ do
...
+    describe "builds correct mde-links for compounds" $ do
```
gromakovsky: It doesn't look like hspec eDSL usage intended by its authors. The string passed to `describe` is supposed to name the thing you are describing and then the string passed to `it` is supposed to contain the behavior you are describing. Like `describe "dog" $ it "loves food" $ do testLovesFood`.

So I think it's more appropriate to write `describe "mde link construction"` here or something like that.

gromakovsky (approved):
Looks good, I have only 3 minor stylistic comments. Btw, why did you amend your first commit instead of pushing a fixup commit?

backend/test/Test/UploadSpec.hs
worm2fed: I wanted to write tests with file upload at first, and then decided to call compound insertion directly and forgot to delete comment)

worm2fed:
>  Btw, why did you amend your first commit instead of pushing a fixup commit?
I've made fixup and then rebase

hm.. it's more convenient to save separate commits before merge to review changes... are

PR #92, opened 2021-05-07
[EDNA-130] Add an optimized getter of filtered experiments 

gromakovsky:
## Description

In order to construct dashboard selectors we use the `/experiments` endpoint which sends a lot of data, especially if you pass no filters (which happens for example if you open dashboard from scratch and it has no filters). So we add another endpoint which returns only data about experiments necessary to construct dashboard selectors. Additionally, we add endpoints to query names of all projects, compounds, etc. because they work faster than `/projects`, `/compounds`, etc.

It's ready, but I want to add more tests/improve existing ones.

## Related issue(s)

Resolves https://issues.serokell.io/issue/EDNA-130

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/backend/src/Edna/DB/Integration.hs
+++ b/backend/src/Edna/DB/Integration.hs
@@ -15,10 +15,12 @@ module Edna.DB.Integration
   , runDeleteReturningList'
   , runSelectReturningOne'
   , runSelectReturningList'
+  , runSelectReturningSet
   ) where
 
 import Universum
 
+import qualified Data.Set as Set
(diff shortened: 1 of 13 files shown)


gromakovsky:
Apparently this PR won't be reviewed/merged any time soon. Maybe we'll come back to it in summer. Here are a couple of things we should improve:
1. `getExperimentsSummary` tests lack tests with more than 1 filter.
2. The second commit has no tests.

gromakovsky:
One more TODO: use new getters of names in this PR or make sure there is an issue about that.

PR #93, opened 2021-05-07
[EDNA-92] Remove unnecessary locale archive

mkaito:
-------

## Description
glibc includes C.UTF-8, which is enough for our purposes. We do not need to
include the entire locale archive.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-92

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/docker.nix
+++ b/docker.nix
@@ -6,7 +6,6 @@
 , backend
 , frontend
 , locale
-, glibcLocales
 , linkFarm
 , runCommand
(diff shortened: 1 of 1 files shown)


gromakovsky (approved):
Apparently it saves about 50 megabytes, looks good. And I've checked that `Trying to connect to the DB…` is printed properly.

PR #95, opened 2021-05-10
[EDNA-133] Add required license information to frontend files

gromakovsky:
## Description

Problem: there are some frontend files that we copy-pasted from
    somewhere. Their licenses require us to include their copyright notice
    and license text, but we don't include them yet. Specifically, these
    files are: `Spinner.scss` style and all PTRootUI fonts.
    
Solution:
1. Refer to the MIT license in `Spinner.scss`.
2. License information about fonts is included into `PTRootUI.scss`
    file where the fonts are imported. It seems to be a common approach
    permitted by the OFL license and.
    
Note that we use `/*!` comments, so they are not removed in production mode.

I am not sure whether providing these links is enough. I can copy whole texts of licenses, I just find it a bit ugly.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-133

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/frontend/src/components/Spinner/Spinner.scss
+++ b/frontend/src/components/Spinner/Spinner.scss
@@ -4,6 +4,15 @@
  * SPDX-License-Identifier: MIT
  */
 
+/*!
+ *  License
+ *  ------------------------------------------------------------------------------
+ *  This code is licensed under the MIT License.
+ *  The license is available at:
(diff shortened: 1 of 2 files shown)


gromakovsky:
@kirelagin please check whether it's good enough or I should also include full texts of licenses. I have checked that comments are present in `.css` that is served in production mode (by downloading frontend image from CI and `docker load`'ing it).

kirelagin (approved):
I did not check that they persist myself, but if they do, then it’s more than enough, IMHO.

PR #96, opened 2021-05-10
Reset button for upload status and upload page improvements

sashasashasasha151:
Problem: we need a way to reset only uploaded file

Solution: after parsing of adding file there is a way now to reset the
file. Also you will see notification with undo. Additionally there are
some upload page style improvements

## Related issue(s)

https://issues.serokell.io/issue/EDNA-122

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/frontend/src/components/UploadStatus/UploadStatus.tsx
+++ b/frontend/src/components/UploadStatus/UploadStatus.tsx
@@ -3,14 +3,20 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
 import React from "react";
-import { useRecoilValue } from "recoil";
-import SuccCheckmark from "../../assets/svg/success-checkmark.svg";
+import { useRecoilState } from "recoil";
+import { Tooltip } from "../Tooltip/Tooltip";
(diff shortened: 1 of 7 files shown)


gromakovsky:
On file reset we show notification that says "You have reset all completed fields". I think it should say "You have reset the file" or something like that, because we reset not all fields.

frontend/src/components/UploadStatus/UploadStatus.tsx
```diff
@@ -28,7 +32,44 @@ export function UploadStatus(): React.ReactElement {
...
+                        >
+                          Undo
+                        </Button>
+                      </div>
+                    ),
+                  });
+                }
+              }}
+            >
+              <Tooltip text="Reset file">
```
gromakovsky: In design this tooltip says "Clear file field", did you change it intentionally? I don't have a preference here btw, I don't know what's better, just pointing out the difference. So I won't object if we keep "Reset file".

gromakovsky (approved):

PR #98, opened 2021-05-10
Invalid sub-experiments info

sashasashasasha151:
Problem: currentry there is no way to see what is wrong with
sub-experiment

Solution: add error tooltip for invalid IC50 and two warnings

## Related issue(s)

https://issues.serokell.io/issue/EDNA-100

- Tests
        silently reappearing again.

- Documentation

### DIFF
--- a/frontend/src/components/IC50Line/IC50Tooltip.tsx
+++ b/frontend/src/components/IC50Line/IC50Tooltip.tsx
@@ -21,7 +21,7 @@ export function IC50Tooltip({ ic50, className }: IC50TooltipProps): React.ReactE
   const errTooltip = (msg: string) => {
     return (
       <Tooltip text={msg} type="error">
-        <span className={cx("ic50value__none", className)} />
+        <span className={cx("ic50value__none", className)}>error</span>
       </Tooltip>
     );
   };
--- a/frontend/src/pages/dashboard/ExperimentsTable/ExperimentsTable.scss
+++ b/frontend/src/pages/dashboard/ExperimentsTable/ExperimentsTable.scss
@@ -36,16 +36,20 @@
     padding: 16 * $px;
   }
 
-  &__subexperiment {
-    margin-bottom: 11 * $px;
-  }
-
   &__subexperiment:nth-child(odd) {
     margin-right: 9 * $px;
   }
 
+  &__subexperiment:nth-child(n + 3):not(.experimentsTable__subexperiment_expanded) {
+    margin-top: 11 * $px;
+  }
+
   &__subexperiment_expanded:nth-child(4n+2) {
     margin-right: 9 * $px;
+  }
+
+  &__subexperiment_expanded:nth-child(n+5) {
+    margin-top: 11 * $px;
   }
 
   &__showCheckbox {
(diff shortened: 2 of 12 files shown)


frontend/src/components/IC50Line/IC50Tooltip.tsx
```diff
@@ -21,7 +21,7 @@ export function IC50Tooltip({ ic50, className }: IC50TooltipProps): React.ReactE
   const errTooltip = (msg: string) => {
     return (
       <Tooltip text={msg} type="error">
-        <span className={cx("ic50value__none", className)} />
+        <span className={cx("ic50value__none", className)}>error</span>
```
gromakovsky, line 24: There was `-` in design for this case, but apparently it was lost. I've asked to update it, you can see it now
image

It's in "Default dashboard+tooltips". But if we change it to `-`, there is a problem because we show tooltip only when cursor is above `-` which is quite small. If you have an idea what to do with it (e. g. show tooltip when cursor is above the whole area where text (numbers) is usually written), let's change `error` to `-` and do it. Otherwise let's just skip this problem and write `error` which is larger.

frontend/src/pages/dashboard/ExperimentsTable/ExperimentsTable.scss
```diff
@@ -36,18 +36,22 @@
...
 
-  &__subexperiment {
-    margin-bottom: 11 * $px;
-  }
-
   &__subexperiment:nth-child(odd) {
     margin-right: 9 * $px;
   }
 
+  &__subexperiment:nth-child(n + 3):not(.experimentsTable__subexperiment_expanded) {
```
gromakovsky, line 43: What do all these expressions in brackets (like `n + 3` here for example) mean?

gromakovsky:
When Recompute fails, we currently show some large modal window with error, it would be better to show this message using our standard notification mechanism instead.

frontend/src/components/IC50Line/IC50Tooltip.tsx
sashasashasasha151: So, as for me `error` is better, because `-` is smth strange for me. I suggest to leave it. As for experiment `-` we don't have it now, will add

frontend/src/pages/dashboard/ExperimentsTable/ExperimentsTable.scss
sashasashasasha151: Some css hacks for not adding addition classes for margins. \
So, `n` mean `0, 1, 2 ...` child. Like `2n` mean each second child, `2n+3` each `3, 5, 7 ...`  child

sashasashasasha151: Childs indexes start from 1 but n indexes from 0 :putya:

gromakovsky: > Some css hacks

Okay, probably I shouldn't delve into it.

gromakovsky (approved):

PR #99, opened 2021-05-11
[EDNA-136] Add more API tests

worm2fed:
## Description

Problem:
We've added some tests for Library API where we call HTTP endpoints. But it's not cover all API and how this API compatible with common standards.

Solution:
Add some integration tests for Dashboard with servant-client and more property tests using servant-quickcheck.

## Related issue(s)

https://issues.serokell.io/issue/EDNA-136

- Tests
        silently reappearing again.

- Documentation


backend/src/Edna/Dashboard/Service.hs
```diff
@@ -61,6 +63,8 @@ makePrimarySubExperiment subExpId = do
 setNameSubExperiment :: SubExperimentId -> Text ->
   Edna (WithId 'SubExperimentId SubExperimentResp)
 setNameSubExperiment subExpId name = do
+  -- Check existence
+  void $ getSubExperiment subExpId
```
gromakovsky: Looks like now we need `transact` here and in `makePrimarySubExperiment`.

gromakovsky (commented):
Regarding "Add existence checks in logic": one additional reason why these checks are useful is that it seems better to return "Invalid ID" error for requests with invalid IDs than to request success and simply do nothing.

backend/test/Test/API/DashboardSpec.hs
```diff
@@ -0,0 +1,263 @@
...
+      { nserName = "newSubExperiment"
+      , nserChanges = one mId
+      }
+
+    aSample :: AnalysisResult
+    aSample = Right Params4PL
+      { p4plA = 29.142616408435742
+      , p4plB = -4.357823061054072
+      , p4plC = 53.239595653095996
+      , p4plD = 1000.3020103924559
```
gromakovsky: It's very fragile to hardcode these numbers because any minor change in Python code will cause these tests to break (AFAIU). I think it's better to call `analyse4PLOne` from test code and compare `AnalysisResult` received from API with the one received from `analyse4PLOne`.

backend/src/Edna/Dashboard/Service.hs
worm2fed: yes, you are right
for now there are two requests

backend/test/Test/API/DashboardSpec.hs
worm2fed: hm.. I do not like this too. Thanks for an advice, I'll do it

backend/test/Test/API/DashboardSpec.hs
```diff
@@ -0,0 +1,263 @@
+-- SPDX-FileCopyrightText: 2021 Serokell <https://serokell.io>
+--
+-- SPDX-License-Identifier: AGPL-3.0-or-later
+
+-- | API testing for Library.
```
gromakovsky: for Dashboard

backend/test/Test/API/DashboardSpec.hs
worm2fed: I decided to not check exact values here, cause we already have such a tests, and here the main goal is check that response is correct, not computation results. I didn't want to duplicate some code from `Test.DashboardSpec`. 
What do you think?

backend/src/Edna/Dashboard/Service.hs
worm2fed: done

backend/test/Test/API/DashboardSpec.hs
worm2fed: done

backend/test/Test/API/DashboardSpec.hs
gromakovsky: Okay, makes sense.

gromakovsky (approved):

